PR #1, opened 2024-09-17
feat(tools): implement move command

argl:
### Description

Implements the `content move` command, along with tests, fixture handling, redirect handling and wiki history updates.

(MP-1530)


crates/rari-doc/src/redirects.rs
```diff
@@ -99,3 +113,721 @@ pub fn resolve_redirect(url: &str) -> Option<Cow<'_, str>> {
...
+                "\t" => "\\t",
+                _ => symbol,
+            };
+            return Err(DocError::InvalidRedirectToURL(format!(
+                "URL '{}' contains forbidden symbol '{}'.",
+                url, escaped_symbol
+            )));
+        }
+    }
+    Ok(())
```
fiji-flo, line 405: Sorry for the drive by comment. I was looking at merge conflicts. This would save the LazyLock (Well that could be replaces by `['\t', '\n']` anyway). 
```suggestion
    if let Some(symbol_index) = url.find(['\t', '\n']) {
        if let Some(escaped_symbol) = &url[symbol_index..]
            .chars()
            .next()
            .map(|c| c.escape_default())
        {
            return Err(DocError::InvalidRedirectToURL(format!(
                "URL '{url}' contains forbidden symbol '{escaped_symbol}'."
            )));
        }
    }
```

fiji-flo (approved):
Great!

PR #4, opened 2024-09-18
feat(locale): add German

caugner:
### Description

Adds the German locale.

### Motivation

We're experimenting with the German locale, and rari doesn't support it yet.


crates/rari-md/src/node_card.rs
```diff
@@ -10,6 +10,9 @@ pub enum NoteCard {
 impl NoteCard {
     pub fn prefix_for_locale(&self, locale: Locale) -> &str {
         match (self, locale) {
+            (Self::Callout, Locale::De) => "Hinweis",
+            (Self::Warning, Locale::De) => "Warnung",
+            (Self::Note, Locale::De) => "Notiz",
```
caugner, line 15:
```suggestion
            (Self::Callout, Locale::De) => "Aufruf:",
            (Self::Warning, Locale::De) => "Warnung:",
            (Self::Note, Locale::De) => "Hinweis:",
```

fiji-flo (approved):
@caugner I added some fixes. Good to go.

PR #24, opened 2024-10-20
feature(cli): sync translated content

argl:
### Description

Implement sync-translated-content for the CLI
(MP-1610)

### Motivation

Ported over from yari


crates/rari-tools/src/sync_translated_content.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    Ok(())
+}
+
+fn md_or_html_exists(slug: &str, locale: Locale) -> Result<bool, ToolError> {
+    let folder_path = root_for_locale(locale)?
+        .join(locale.as_folder_str())
+        .join(slug.to_lowercase());
+    let md_path = folder_path.join("index.md");
+    let html_path = folder_path.join("index.html");
```
caugner, line 329: We no longer need to support `index.html` (cf. https://github.com/mdn/yari/pull/7575).

crates/rari-types/src/locale.rs
```diff
@@ -112,6 +112,24 @@ static LOCALES_FOR_GENERICS_AND_SPAS: LazyLock<Vec<Locale>> = LazyLock::new(|| {
...
+    let translated_locales = [
+        Locale::Es,
+        Locale::Fr,
+        Locale::Ja,
+        Locale::Ko,
+        Locale::PtBr,
+        Locale::Ru,
+        Locale::ZhCn,
+        Locale::ZhTw,
+    ];
```
caugner, line 125:
```suggestion
    let translated_locales = [
        Locale::De,
        Locale::Es,
        Locale::Fr,
        Locale::Ja,
        Locale::Ko,
        Locale::PtBr,
        Locale::Ru,
        Locale::ZhCn,
        Locale::ZhTw,
    ];
```

caugner (changes requested):
Only a quick review that doesn't replace @fiji-flo's.

crates/rari-types/src/locale.rs
argl: Agreed, I was not aware that `de` is now officially supported

crates/rari-tools/src/sync_translated_content.rs
argl: Ok, I'll remove that

fiji-flo (approved):
Looks good.

PR #26, opened 2024-10-23
chore: update authors

caugner:
### Description

Sets the MDN Engineering Team as the author.


fiji-flo:
Changed via https://github.com/mdn/rari/commit/df407d3e284024e3384e4547013100ac5cb3c8a9
We'll reflect the best point of contact in the authors field and we'll update it as knowledge/ownership/responsibility change.

caugner:
> We'll reflect the best point of contact in the authors field

@fiji-flo May I ask why you consider your personal email address the best point of contact, given that you're a member and owner of the mdn-dev group, and should therefore receive emails sent to that group? My understanding is that we should use our work email address here for legal reasons, to avoid giving the impression that you authored the software outside of work.

argl (approved):

argl (approved):

argl (approved):

PR #28, opened 2024-10-25
chore(doc/pages): consider de repo

caugner:
### Description

Uses the translated-content-de repo to construct github/commit urls for de.

### Motivation

The German locale has a different repo.


fiji-flo (approved):
Let's add this for now.

PR #31, opened 2024-10-29
Mp 1696 diff html massage

argl:
### Description

pre / code logic + inswertion on rewriter leve, introduction of lol_html on diff-test
(MP-1696)

### Motivation

yari/rari parity


fiji-flo (approved):
Great

PR #32, opened 2024-10-30
chore(diff): add allowlist support for diffing content

argl:
### Description

Allowlist of <filepath> <jsponpath> pairs for diffing

### Motivation

rari/yari parity diffing support


fiji-flo (approved):
👍

PR #34, opened 2024-11-01
feat(html): post process dts

fiji-flo:
### Description

This moves the insert links logic for dts to post processing.

### Motivation

support e.g.

```
- *foo* bar
  - : 2000
```

so that `*foo* bar` is a link and not only `*foo*`.


argl (approved):
LGTM

PR #70, opened 2024-12-27
fix(l10n): correct the repo name to translated content

yin1999:
### Description

The repo name of translated content is `mdn/translated-content`, not `mdn/translated_content`.

### Motivation

Check the "Recent contributions" section in the [Home page](https://developer.mozilla.org/). The 
URLs to translated-content's PRs are broken. It's also reported in the MDN official Discord channel.


fiji-flo (approved):
Thanks for fixing

PR #72, opened 2025-01-03
feat(sidebars/jsref): support temporal

Josh-Cena:
As a part of https://github.com/mdn/content/pull/37344/, we need to support displaying `Temporal` in a similar fashion as `Intl`. To do so I made several refactors to avoid repeating myself.

As an aside I'm amused by how much we use `Cow` when most of the strings here are immutable anyway.


fiji-flo:
@Josh-Cena I opened a PR against your branch, can you take a look: https://github.com/Josh-Cena/rari/pull/1

Issue #73, opened 2024-12-25
CSSRef is missing functions with nested slugs


skyclouds2001:
### MDN URL

.

### What information was incorrect, unhelpful, or incomplete?

same as title

e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix

note that in my memory, they are in sidebar a fewer times ago

### What did you expect to see?

these css functions are in sidebar generated by {{CSSRef}

### Do you have anything more you want to share?

maybe this is related with https://github.com/mdn/rari

caugner:
The Glossary sidebar is also affected, see https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility, which is missing in the sidebar.

fiji-flo:
Reopen till I have a content PR

PR #74, opened 2025-01-03
feat(sidebars): add depth option to listSubPages[Grouped]

caugner:
### Description

Allows to `listSubPages` recursively.

### Motivation

Some CSS functions like [color-mix()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix) are nested, so they currently no longer appear in the sidebar, because `depth` defaults to 1.

### Additional details

To test this, apply this change in content:

```diff
diff --git a/files/sidebars/cssref.yaml b/files/sidebars/cssref.yaml
index 3f2b20e1f8..a6b0e741d6 100644
--- a/files/sidebars/cssref.yaml
+++ b/files/sidebars/cssref.yaml
@@ -96,6 +96,7 @@ sidebar:
     path: /Web/CSS
     title: Functions
     tags: css-function
+    depth: 2
     details: closed
   - type: listSubPages
     path: /Web/CSS
```

Fixes https://github.com/mdn/rari/issues/73.

### DIFF
--- a/crates/rari-doc/src/helpers/subpages.rs
+++ b/crates/rari-doc/src/helpers/subpages.rs
@@ -189,6 +189,7 @@ pub fn list_sub_pages_grouped_internal(
     out: &mut String,
     url: &str,
     locale: Locale,
+    depth: Option<usize>,
     ListSubPagesContext {
         sorter,
         page_types,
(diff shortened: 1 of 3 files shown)


caugner:
Unfortunately, this approach doesn't work, because the `list_subpages_internal()` implementation currently always renders a tree structure, and skips intermediate nodes that don't match the filtered page type.

It doesn't seem trivial to extend the behavior of this function, because it mixes querying and displaying, e.g. sub-trees are individually sorted, so we cannot easily display nested pages in a flat list.

PR #81, opened 2025-01-09
fix(l10n): correct the prefix string for notecards in zh-TW locale

yin1999:
### Description

correct the prefix string for notecards in zh-TW locale.

### Motivation

The prefix string for ntoecards in zh-TW locale appears to be identical to the zh-CN locale. Fix them by copying those prefix strings from [yari](https://github.com/mdn/yari/blob/0b5276be8dc2b10992f1fe91ebdc8f7c0285430e/markdown/localizations/zh-TW.json)

### Additional details

A screenshot from https://developer.mozilla.org/zh-TW/docs/Web/CSS/@font-face#%E6%8F%8F%E8%BF%B0:

image

/cc @mdn/yari-content-zh


fiji-flo:
Thanks and sorry for the mistake. I think we should eventually move these to content.

fiji-flo (approved):

PR #88, opened 2025-01-18
fix(build/parser): parse empty string args for macros as `None`s

yin1999:
### Description

replace empty string args for macos with `None`s

### Motivation

In mdn/content, we are used to use empty string to represent the argument is not provided, which should be `None` instead of `Some(EmptyString)`. But we have use some judgement like `if let Some(anchor) = anchor` to check if the string argument is provided. This will evaluate to `true` as the judgement encounters a string.

I found an incorrect case in `/en-US/docs/Mozilla/Add-ons/WebExtensions/API/pageAction/onClicked` page:

image

There is a hash symbol at the end of the link text.

I think the easiest solution is to replace the empty string args with `None`s. As we use similar logic to the above to determine whether the parameters are provided in many macos. Some cases (Maybe not all of them):

- crates/rari-doc/src/templ/templs/links/domxref.rs (we have checked whether the arg is an empty string)
- crates/rari-doc/src/templ/templs/links/htmlxref.rs
- crates/rari-doc/src/templ/templs/links/http.rs
- crates/rari-doc/src/templ/templs/links/jsxref.rs
- crates/rari-doc/src/templ/templs/links/webextapixref.rs (the one I found the issue)

I haven't fully tested this change, so I'm not sure if it will break any pages. If you have any ideas, please let me know.

### DIFF
--- a/crates/rari-doc/src/templ/parser.rs
+++ b/crates/rari-doc/src/templ/parser.rs
@@ -76,6 +76,7 @@ fn to_arg(pair: Pair<'_, Rule>) -> Option<Arg> {
         Rule::single_quoted_string => pair.into_inner().next().and_then(to_arg),
         Rule::double_quoted_string => pair.into_inner().next().and_then(to_arg),
         Rule::backquoted_quoted_string => pair.into_inner().next().and_then(to_arg),
+        Rule::empty_string => None,
         Rule::sq_string => {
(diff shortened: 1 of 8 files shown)


fiji-flo:
Thanks for this. I opened a PR https://github.com/yin1999/rari/pull/1 to move the logic into the actual parser. Please take a look and merge.

PR #90, opened 2025-01-21
fix(html): trim the first empty line in `<pre>` tag

yin1999:
### Description

trim the first empty line in `<pre>` tag

### Motivation

The adding of `<code>` tag makes the first empty line would not be ignored by User Agent, see the [spec](https://www.w3.org/MarkUp/1995-archive/Elements/PRE.html). Ref: https://github.com/mdn/yari/issues/12364#issuecomment-2561133409

So trimming the first empty line in `<pre>` tag to simulate the semantics.

Fixes: mdn/yari#12364

Test: check the rendered html in `/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action`:

Before:

image

After:

image

### DIFF
--- a/crates/rari-doc/src/html/rewriter.rs
+++ b/crates/rari-doc/src/html/rewriter.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::collections::HashSet;
 
 use lol_html::html_content::ContentType;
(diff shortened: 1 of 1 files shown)


fiji-flo:
Great thanks for this I added a minor fix in https://github.com/yin1999/rari/pull/2

Issue #89, opened 2025-01-20
"Error: could not find syntax for this item" on CSS repeat() page


BANOnotIT:
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#formal_syntax

### What information was incorrect, unhelpful, or incomplete?

Section shows "Error: could not find syntax for this item", which seems to be system error

### What did you expect to see?

I think people should not see system errors. Section should be hidden if there's nothing to show there

fiji-flo:
@pepelsbey yes that's the case. https://drafts.csswg.org/css-grid-2/#repeat-syntax:

> The generic form of the [repeat()](https://drafts.csswg.org/css-grid-2/#funcdef-repeat) syntax is, approximately,
> 
> repeat( [ [<integer [1,∞]>](https://drafts.csswg.org/css-values-4/#integer-value) [|](https://drafts.csswg.org/css-values-4/#comb-one) auto-fill [|](https://drafts.csswg.org/css-values-4/#comb-one) auto-fit ] [,](https://drafts.csswg.org/css-values-4/#comb-comma) [<track-list>](https://drafts.csswg.org/css-grid-2/#typedef-track-list) )

Ideally we would fix this upstream but I don't know if that's possible.

An solution is to display the syntax for [track-repeat](https://drafts.csswg.org/css-grid-2/#typedef-track-repeat) and the others (which needs tweeking the csssyntax macro 😮‍💨) or we add dedicated pages for those an link there.

BANOnotIT:
The issue is still present on MDN page. Will it be fixed automatically later?

fiji-flo:
Sorry this was automatically closed. 

The macro on https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#formal_syntax must be changed to.

`{{CSSSyntaxRaw("<track-repeat> <auto-repeat> <fixed-repeat> <name-repeat>")}}`

caugner:
@BANOnotIT Would you be interested in opening a content PR to update [this line](https://github.com/mdn/content/blob/ecd91b52ed14e4c45a245928c7202a8a265b47a3/files/en-us/web/css/repeat/index.md#L123)?

caugner:
It looks like angle brackets get replaced by HTML entities, resulting in this content line:

```md
{{CSSSyntaxRaw("&lt;track-repeat&gt; &lt;auto-repeat&gt; &lt;fixed-repeat&gt; &lt;name-repeat&gt;")}}
```

[...]

PR #99, opened 2025-01-24
fix(macros/CSSSyntaxRaw): decode entities in parameter

caugner:
### Description

Decodes entities in the parameter passed to the new `CSSSyntaxRaw` macro.

### Motivation

The markdown inter in content requires `<` and `>` to be converted to HTML entities, so we need to decode these first.

### Additional details

Tested locally:

1. Ran `cargo run serve` on this branch.
2. Ran `yarn start:rari-external` in yari.
3. Checked out https://github.com/mdn/content/pull/37787 in content.
4. Looked at http://localhost:5042/en-US/docs/Web/CSS/repeat#formal_syntax

image

Relates to https://github.com/mdn/rari/issues/89.


caugner:
Closing, as the solution is to use backticks instead of quotes.

PR #126, opened 2025-02-07
feat(tools): remove redirects that reference to deleted docs

yin1999:
### Description

remove redirects that reference to deleted docs.

### Motivation

To be written


fiji-flo (approved):
good work! thanks

PR #136, opened 2025-02-25
fix(macro): fix some CSS redirects

OnkarRuikar:
### Description

Update URLs in CSS macros


fiji-flo (approved):
Thanks that's great. I'll do a release tomorrow.

May we should think about moving some of these URLs in got jsondata?

PR #141, opened 2025-03-10
ci(dependabot): setup for GitHub Actions

caugner:
### Description

Enables weekly Dependabot runs for GitHub Actions.

### Motivation

We don't want to miss important updates to these.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


caugner (commented):
Self-approval.

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "build(deps): "
```
caugner, line 8:
```suggestion
      prefix: "ci(deps): "
```

Issue #140, opened 2025-03-06
Inappropriate `_static` suffix on some static methods and properties in sidebar


webstrand:
### What information was incorrect, unhelpful, or incomplete?

The sidebar on <https://developer.mozilla.org/en-US/docs/Web/API/File_API> lists two methods at the bottom [URL.createObjectURL_static](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static)
    [URL.revokeObjectURL_static](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL_static) giving them incorrect names. The `_static` suffix is invalid.

This seems to affect other sidebars, too. For example <https://developer.mozilla.org/en-US/docs/Web/API/CSS_Painting_API> and <https://developer.mozilla.org/en-US/docs/Web/API/Force_Touch_events>

### What did you expect to see?

No `_static` suffix

### Do you have any supporting links, references, or citations?

This might be the data that controls the sidebar generation <https://github.com/mdn/content/blob/8379a43c93b36c69c44a4a0b23219aeae0affb0f/files/jsondata/GroupData.json#L555>

### Do you have anything more you want to share?

(Also your issue report form [CONTRIBUTING.md](https://github.com/mdn/blob/main/CONTRIBUTING.md) link is dead)

caugner:
I can reproduce the issue, but it does not seem to be a regression, as the sidebar already contained this error [in November 2024](https://web.archive.org/web/20241122080711/https://developer.mozilla.org/en-US/docs/Web/API/File_API/) before we migrated to rari.

PR #145, opened 2025-03-12
fix(sidebars): strip "_static" suffix

caugner:
### Description

Strips the `_static` suffix from page titles displayed in the `APIRef` and `DefaultAPISidebar` sidebars.

### Motivation

The suffixes are confusing, as they aren't part of the API name.

Fixes https://github.com/mdn/rari/issues/140.


fiji-flo (approved):
👍

PR #146, opened 2025-03-12
fix(homepage): use text content for summary on homepage

fiji-flo:
### Description

The summary for the homepage should be text rather than html.

### DIFF
--- a/crates/rari-doc/src/helpers/summary_hack.rs
+++ b/crates/rari-doc/src/helpers/summary_hack.rs
@@ -1,5 +1,6 @@
 use itertools::Itertools;
 use rari_md::{m2h_internal, M2HOptions};
+use scraper::Html;
 
 use crate::error::DocError;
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
lgtm!

crates/rari-doc/src/helpers/summary_hack.rs
```diff
@@ -46,3 +47,8 @@ pub fn strip_paragraph_unchecked(input: &str) -> &str {
 
     out
 }
+
+pub fn text_content(html_str: &str) -> String {
```
caugner, line 51: Nit: Maybe `html_to_text()`

```suggestion
pub fn html_to_text(html_str: &str) -> String {
```

fiji-flo: it comes from https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

PR #156, opened 2025-03-14
fix: update paths for SVG/MathML section move

pepelsbey:
### Description

We've recently moved some docs in MathML and SVG sections.
This updates some paths so we don't have unnecessary redirects.

Depends one these PRs:

- [x] https://github.com/mdn/content/pull/38379
- [x] https://github.com/mdn/content/pull/38589


caugner (approved):
LGTM, but marking as draft to avoid accidental merge while the SVG content PR is not yet merged.

PR #157, opened 2025-03-14
fix(notecard): workaround for prettier

fiji-flo:
Support

```
> [!WARNING]\
> [Prettier](https://prettier.io) will break this.
```

> [!WARNING]\
> [Prettier](https://prettier.io) will break this.

to render as inline warning.


yin1999:
We may not need this workaround anymore, and we have replaced all the old notecard style with GFM alerts syntax in mdn/content: mdn/content#40056

caugner:
> We may not need this workaround anymore

Based on this, I'm closing this PR. We can reopen if there is evidence that this is still needed.

PR #180, opened 2025-04-01
feat(json): add renderer to index.json for fred

fiji-flo:
### Description

Add top level `renderer` field to index.json for fred to use for routing.

### DIFF
--- a/crates/rari-doc/src/pages/templates.rs
+++ b/crates/rari-doc/src/pages/templates.rs
@@ -1,0 +1,137 @@
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use super::json::{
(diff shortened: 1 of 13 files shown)


crates/rari-doc/src/pages/templates.rs
```diff
@@ -0,0 +1,137 @@
...
+    SpaPlusUpdates,
+    SpaPlusSettings,
+    SpaPlusAiHelp,
+    SpaPlay,
+    SpaSearch,
+}
+
+#[derive(Debug, Clone, Serialize, JsonSchema)]
+#[serde(tag = "renderer")]
+pub enum SpaPage {
```
fiji-flo, line 31: This might move to a macro, but getting the outer API out is more important right now.

PR #195, opened 2025-04-15
feat(sidebars): allow webgl-extension-method to show up in ApiRef sidebar

Josh-Cena:
Currently, the extension methods are not available in the sidebar:

image

By treating `webgl-extension-method` the same as `web-api-instance-method`, this page can now be included in the sidebar.


fiji-flo (approved):
👍

PR #200, opened 2025-04-25
feat(baseline): display baseline by bcd key

fiji-flo:
### Description

Use the bcd key to display the baseline banner. Now we display the status for the bcd key of the page instead of the feature status. We also start showing the baseline status if the feature status differs from the key status.


LeoMcA (approved):
nice, looks good! thanks for inverting the asterisk logic, and with the bugfix for subkeys, the code is nice and clear now

PR #207, opened 2025-04-30
fix(md): render DD in order

### LINKED ISSUE #190: Don't render double `- :` as DD, or render them in order

Try the following Markdown:

```md
- foo
  - : item1
  - : item2
```

Currently, it's rendered as a DL with two DDs. This may or may not be intended because it's not documented. But what's bizarre is that "item 2" shows up before "item 1".

Either render it as a plain UL like before (which will be reported by my automation), or render the two DDs in order.

_No response_

_No response_

cheeslide:
### Description

- reverse the order of processing
- add corresponding test

### Motivation

It is a bug.

### Additional details

None.

fix #190

### DIFF
--- a/crates/rari-md/src/dl.rs
+++ b/crates/rari-md/src/dl.rs
@@ -35,7 +35,7 @@ pub(crate) fn convert_dl<'a>(list: &'a AstNode<'a>) {
             continue;
         }
         last_child.detach();
-        for item in last_child.children() {
+        for item in last_child.reverse_children() {
             if let Some(i) = item.first_child() {
                 if !matches!(i.data.borrow().value, NodeValue::Paragraph) {
                     break;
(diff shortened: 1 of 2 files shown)


fiji-flo (changes requested):
Thanks, that's great. One tiny change and we're ready to go.

cheeslide:
Thanks for your timed reply.

Do you mean I need to sign all the commits, or just the last commit?

fiji-flo:
Yes, we enforce all commits to be signed. Thanks a lot 🙏 

image

cheeslide:
@fiji-flo Is it ok?

Josh-Cena:
@fiji-flo: so is multiple-DD expected behavior now? In this case should we update https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#definition_lists?

PR #231, opened 2025-06-10
fix(css-syntax): trim trailing line breaks

### LINKED ISSUE #229: Formal syntax produces two trailing empty newlines

The Formal syntax on https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/rect#formal_syntax contains two lines with syntax, followed by two empty lines:

Image

I would expect to only see the two lines with syntax:

Image

These `<br/>` appear to be written here:

https://github.com/mdn/rari/blob/d06ef765f10141101153e1224a8ca909d2e251ea/crates/css-syntax/src/syntax.rs#L824-L827

https://github.com/mdn/rari/blob/d06ef765f10141101153e1224a8ca909d2e251ea/crates/css-syntax/src/syntax.rs#L445-L448

_No response_

caugner:
### Description

Removes trailing line breaks in the CSSSyntax macro output.

Fixes https://github.com/mdn/rari/issues/229.

### DIFF
--- a/crates/css-syntax/src/syntax.rs
+++ b/crates/css-syntax/src/syntax.rs
@@ -821,10 +821,12 @@ fn write_formal_syntax_internal(
     write!(out, r#"<pre class="notranslate css-formal-syntax">"#)?;
     let mut constituents = renderer.get_constituent_syntaxes(syntax)?;
 
+    let mut constituents_out = String::new();
     for constituent in constituents.iter().skip(if skip_first { 1 } else { 0 }) {
-        renderer.render(&mut out, constituent)?;
-        out.push_str("<br/>");
+        renderer.render(&mut constituents_out, constituent)?;
+        constituents_out.push_str("<br/>");
     }
+    out.push_str(constituents_out.trim_end_matches("<br/>").trim_end());
 
     let specs = constituents.iter_mut().fold(vec![], |mut acc, s| {
         if let Some(spec) = s.specs.take() {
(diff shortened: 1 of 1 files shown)


crates/css-syntax/src/syntax.rs
```diff
@@ -821,10 +821,12 @@ fn write_formal_syntax_internal(
...
     let mut constituents = renderer.get_constituent_syntaxes(syntax)?;
 
+    let mut constituents_out = String::new();
     for constituent in constituents.iter().skip(if skip_first { 1 } else { 0 }) {
-        renderer.render(&mut out, constituent)?;
-        out.push_str("<br/>");
+        renderer.render(&mut constituents_out, constituent)?;
+        constituents_out.push_str("<br/>");
     }
+    out.push_str(constituents_out.trim_end_matches("<br/>").trim_end());
```
fiji-flo, line 829:
```rust
    for (i, constituent) in constituents
        .iter()
        .skip(if skip_first { 1 } else { 0 })
        .enumerate()
    {
        if i > 0 {
            out.push_str("<br/>");
        }
        renderer.render(&mut out, constituent)?;
    }
```

This would do the trick but leaves the last `<br/>` but I think keeping it is better.

caugner: What is the purpose of the trailing `<br/>`?

fiji-flo: Sorry, there is no purpose. Most of our formatters add a trailing new line it has ups and downs. But I really see no reason why this matters here?

caugner: Okay, changes applied.

fiji-flo (approved):

PR #239, opened 2025-06-14
feat(client): add GitHub token support for API requests

yin1999:
### Description

Add GitHub token support for API requests, this is a part of fix to resolve the problem described on discord (Got `Error: invalid type: map, expected a sequence` when running [yarn build in GitHub workflows](https://github.com/mdn/content/blob/main/.github/workflows/pr-test.yml#L109)):

image

I finally figure out what is going on: The error was not caused by "updating xyz To va.b.c", but by rari encountering a rate limiting problem when trying to [get the release info list on GitHub using GitHub API](https://github.com/mdn/rari/blob/82b80b07c462d047648c2c36a242caa8163d5ba5/crates/rari-deps/src/github_release.rs#L21-L26).

To verify this, you can remove the cached version check info, set the wrong **GITHUB_TOKEN** in the environment variable, and you will get output similar to the following:

image

> Note: I am not sure whether we need to do similar authorization for requests to "github.com". Currently, github's documentation does not describe a rate limit policy for github.com.

### Motivation

Our CI for building preview documents occasionally fails to build documents in the same way as above.

### DIFF
--- a/crates/rari-deps/src/client.rs
+++ b/crates/rari-deps/src/client.rs
@@ -1,9 +1,38 @@
 use reqwest::blocking::Response;
+use url::{Url, Host};
 
 pub fn get(url: impl AsRef<str>) -> reqwest::Result<Response> {
-    reqwest::blocking::ClientBuilder::new()
-        .user_agent("mdn/rari")
-        .build()?
+    let mut client = reqwest::blocking::ClientBuilder::new().user_agent("mdn/rari");
+
+    // check if the URL's host is api.github.com
+    if is_github_url(url.as_ref()) {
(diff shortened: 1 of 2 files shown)


fiji-flo:
Thanks a lot. I PR'd a minor change https://github.com/yin1999/rari/pull/3 but really great.

fiji-flo (approved):

yin1999:
> I'll make a release tomorrow. There's one change in this release that I don't want to do on a Sunday evening 😅 

It doesn't matter. The issue fixed by this PR does not occur often.

PR #262, opened 2025-06-30
fix: only return a single parent item for blog index parents/breadcrumbs

argl:
### Description

Only return a single paretn item for the blog index, regardless of the page we are on

### Motivation

Multiple breadcrumbs were created on pages other than the base page/url


caugner (commented):
Can we separate the clippy changes into a separate PR, or make this PR only about the clippy fixes?

Issue #273, opened 2025-07-09
Community page has title "Contribute to MDN | Contribute to MDN" and breadcrumb "Contribute to MDN"


caugner:
### What information was incorrect, unhelpful, or incomplete?

The [Community page](https://fred.review.mdn.allizom.net/en-US/community) has an unexpected title and breadcrumb:

- `<title>`  is `Contribute to MDN | Contribute to MDN`
- breadcrumb is `Contribute to MDN`

### What did you expect to see?

Expected:

- `<title>` to be `Contribute to MDN | MDN`
- breadcrumb to be `MDN Community`

caugner:
Fixed via https://github.com/mdn/rari/pull/274 and https://github.com/mdn/generic-content/pull/14.

PR #275, opened 2025-07-09
enhance(pages): default title-suffix to "MDN"

caugner:
### Description

Adds a suffix "MDN" to all generic pages without a specific one.

### Motivation

We don't want to specify these explicitly for each page.

Currently, the About page's title is "About MDN", whereas the Community's page is "Contribute to MDN | Contribute to MDN".

Part of https://github.com/mdn/rari/issues/273.

### DIFF
--- a/crates/rari-doc/src/pages/build.rs
+++ b/crates/rari-doc/src/pages/build.rs
@@ -383,11 +383,13 @@ fn build_generic_page(page: &Generic) -> Result<BuiltPage, DocError> {
             toc,
         },
         short_title: page.meta.short_title.clone(),
-        page_title: concat_strs!(
-            page.meta.title.as_str(),
-            " | ",
-            &page.meta.title_suffix.as_deref().unwrap_or("MDN")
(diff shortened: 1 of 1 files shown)


crates/rari-doc/src/pages/build.rs
```diff
@@ -379,11 +379,7 @@ fn build_generic_page(page: &Generic) -> Result<BuiltPage, DocError> {
             title: page.meta.title.clone(),
             toc,
         },
-        page_title: if let Some(suffix) = &page.meta.title_suffix {
-            concat_strs!(page.meta.title.as_str(), " | ", suffix)
-        } else {
-            page.meta.title.clone()
-        },
+        page_title: concat_strs!(page.meta.title.as_str(), " | ", &page.meta.title_suffix.as_deref().unwrap_or("MDN")),
```
caugner, line 382: Hm, should we avoid the MDN suffix if the title already includes "MDN"?

PR #311, opened 2025-09-15
feat(baseline): include data for discouraged features

LeoMcA:
Part of: https://github.com/mdn/fred/issues/654

Allows data for discouraged features to be included in our index.jsons, while maintaining the same asterisk logic as before: a discouraged feature is treated as having no baseline status. Without the asterisk logic change, the [font page](https://developer.mozilla.org/en-US/docs/Web/CSS/font) loses its asterisk.


crates/rari-data/src/feature_urls.rs
```diff
@@ -0,0 +1,44 @@
+use indexmap::IndexMap;
+use std::sync::LazyLock;
+
+static FEATURE_MDN_URLS: LazyLock<IndexMap<&'static str, &'static str>> = LazyLock::new(|| {
+    [
+        ("async-clipboard", "/docs/Web/API/Clipboard_API"),
```
caugner: Would be nice if web-features could add an `mdn` field: https://github.com/web-platform-dx/web-features/blob/96d7e3c4e91500cf84f31cbab590f539566980ae/features/async-clipboard.yml#L1-L5

LeoMcA: We've been talking about how best to maintain the web-feature <--> mdn mapping: once we've decided (and implemented) that, I'm sure we'll be able to add an mdn field to web-features.

Perhaps for these "features pointed at by discouraged features" it makes sense to add them manually in the web-features repo though

crates/rari-data/src/baseline.rs
```diff
@@ -185,9 +202,38 @@ impl WebFeatures {
...
+                                        alternative
+                                    );
+                                    None
+                                }
+                            },
+                        )
+                    })
+                    .collect(),
+                None => Vec::new(),
+            };
```
globau, line 230: `self.feature_data_by_name(alternative)` returns `None` when the alternative key doesn't exist in the web-features data - we should log a warning in this case.  Current logic won't generate any warnings.

globau (approved):
lgtm, just a minor missing warning

PR #317, opened 2025-09-18
feat(css-syntax): migrate to @webref/css 7

### LINKED ISSUE #304: Migrate to @webref/css v7
See https://github.com/w3c/webref/issues/1519#issuecomment-2958966410 for details.

Currently, this causes build warnings, and breaks the CSSSyntax macro:

```
Updating @webref/css to 7.0.11

thread 'main' panicked at crates/rari-deps/src/webref_css.rs:70:63:
called `Result::unwrap()` on an `Err` value: Error("invalid type: null, expected a sequence", line: 0, column: 0)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

See also: https://unpkg.com/@webref/css@7.0.11/CHANGELOG.md

> ## v7.0.11 - 2025-09-04
> 
> CSS extracts are now consolidated into a single file.
> 
> ### Breaking changes
> 
> - A single consolidated `css.json` file replaces CSS extracts per spec.
> 
> ### How to upgrade
> 
> In version 6, `listAll()` resolved with an object where the keys were spec shortnames, and the values mostly matched the structure of the consolidated object that the function now returns. On top of the consolidation itself, main differences are:
> - Functions and types were merged in a `values` category in version 6, they now appear in separate `functions` and `types` categories.
> - Functions and types that were scoped to another CSS feature appeared nested under that CSS feature in version 6. They now appear directly under `functions` and `types` with a `for` key that contains the list of scoping features for that feature. A scoping feature may be a property, a function or a type. When the scoping feature is a type, its name in the `for` key is enclosed between `<` and `>`.
> - Feature syntaxes were stored in a `value` key. That key is now named `syntax`. Actual syntax values are the same as before and can be parsed with CSSTree.
> - The shortname of the spec (or specs) that defines a feature is no longer readily available (but note the `href` key targets the defining spec). If you need the specs' shortnames, please [raise an issue](https://github.com/w3c/webref/issues) to describe your use case.
> - The consolidation removes duplicates, merging extended definitions into a single feature. The definition from the latest spec level is used when a feature is defined in more than one level. If you need the definition from earlier levels, please [raise an issue](https://github.com/w3c/webref/issues) to describe your needs!
> - Some of the possible values that a CSS feature could take appeared nested under that feature definition in a `values` key in version 6. Such values are no longer reported in the new version: they were confusing in any case because they did not cover the whole range of values that a feature could take, and could contain values that were not atomic keyword values. Values could be re-introduced when CSS specs are more systematic about them. In the meantime, you will need to parse the feature's syntax (the `syntax` key) to extract keyword values.
> 
> Additionally, the package now also exposes a new `index()` async method that resolves with an object similar to that returned by `listAll()`, except that lists of features under each category are indexed by feature names. For scoped features, the feature name is used as identifier when it is unambiguous. If a feature is defined differently for different scopes, the feature name is completed with the name of the first scoping feature to disambiguate. For example, `type() for @function`, `type() for attr()`.

argl:
### Description

This implements the usage of data from [webref-css](https://github.com/w3c/webref/tree/main) version 7.x. That version only provides a consolidated version of the data, so we can get rid of the brittle logic around selecting a CSS level we show. The editorial responsibility therefore lies with the maintainers of webref-css.

This is the basis of our CSS formal syntax rendering.

### Motivation

Upstream major version change.

### Additional details

[Webref-css changelog](https://github.com/w3c/webref/blob/main/packages/css/CHANGELOG.md)
[Diff of the document changes induced by this PR](https://gist.github.com/argl/044a135c2a75ea324cac99348e6e91d7): 255 changes

#### `<color>` example formal syntax
Before
image

After
image

There are numerous additions to this in the v7 data, namely the correct introduction of `light-dark()` (which we document on the page) and `device-cmyk()` (which is not implemented by any browser by time of writing).

Since the specification titles are no more included with the webref data, titles are pulled in from the [browser-specs](https://www.npmjs.com/package/browser-specs) data package.

Fixes #304


argl:
Waiting for a decision from content for the copy, style and placement for a formal syntax disclaimer.

argl:
Content decision is done, a [PR](https://github.com/mdn/content/pull/41344) adding the necessary translated text fragments is under review.

crates/css-syntax-types/src/lib.rs
```diff
@@ -45,48 +43,92 @@ impl From<&Css> for Css {
...
 impl From<&AtRule> for AtRule {
     fn from(value: &AtRule) -> Self {
         value.clone()
     }
 }
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct Function {
+    #[serde(rename = "specLink", default, skip_serializing_if = "Option::is_none")]
+    pub spec_link: Option<SpecLink>,
```
caugner, line 66: The order is different to Type, is that intentional? It would be nice to be somewhat consistent, assuming the order is relevant for serialisation.

crates/css-syntax-types/src/lib.rs
```diff
@@ -45,48 +43,92 @@ impl From<&Css> for Css {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(deny_unknown_fields)]
```
caugner, line 48: Why do we start allowing unknown fields?

crates/css-syntax-types/src/lib.rs
```diff
@@ -45,48 +43,92 @@ impl From<&Css> for Css {
...
 pub struct AtRule {
```
caugner, line 46: Given that all these types share some fields, could these be extracted to a common base type, to ensure consistency?

crates/css-syntax-types/src/lib.rs
```diff
@@ -533,17 +578,32 @@ impl<'de> serde::Deserialize<'de> for Shortname {
...
     pub url: Url,
 }
-impl From<&SpecInExtract> for SpecInExtract {
-    fn from(value: &SpecInExtract) -> Self {
+impl From<&SpecLink> for SpecLink {
+    fn from(value: &SpecLink) -> Self {
         value.clone()
     }
 }
+// In order to use it with in a BtreeSet<SpecLink>, implement `PartialOrd` and `Ord`
```
caugner, line 592:
```suggestion

// In order to use it with in a BtreeSet<SpecLink>, implement `PartialOrd` and `Ord`
```

crates/css-syntax-types/src/lib.rs
caugner, line 1: This file is getting very long, could some parts be extracted, e.g. to webref_css_types?

crates/rari-deps/src/webref_css.rs
```diff
@@ -22,81 +22,218 @@ fn by_name(values: Vec<Value>) -> BTreeMap<String, Value> {
...
 pub fn update_webref_css(base_path: &Path) -> Result<(), DepsError> {
-    if let Some(package_path) = get_package(
-        "@webref/css",
-        &deps()
-            .webref_css
-            .clone()
-            .or_else(|| Some(VersionReq::parse(">=6.0.0, <7.0.0").unwrap())),
-        base_path,
-    )? {
+    let url_to_title = if let Some(browser_specs_path) =
```
caugner, line 177: Could this url-titles bit be extracted to a separate function?

crates/rari-types/src/settings.rs
```diff
@@ -20,6 +20,8 @@ pub struct Deps {
     pub web_specs: Option<VersionReq>,
     #[serde(alias = "@webref/css")]
     pub webref_css: Option<VersionReq>,
+    #[serde(alias = "browser-specs")]
+    pub browser_specs: Option<VersionReq>,
```
caugner, line 24: Move above for alphabetical order.

crates/rari-doc/src/templ/templs/csssyntax.rs
```diff
@@ -67,12 +67,14 @@ pub fn csssyntax(name: Option<String>) -> Result<String, DocError> {
         ),
         &TOOLTIPS,
         Some(sources_prefix),
-    )?)
+    )?;
+
+    Ok(out)
```
caugner, line 72: Nit: Revert?

crates/rari-deps/src/webref_css.rs
```diff
@@ -22,81 +22,218 @@ fn by_name(values: Vec<Value>) -> BTreeMap<String, Value> {
...
+        } else {
+            println!("could not read url-titles.json, link title fields will not be populated");
+            BTreeMap::new()
+        }
+    }
+
+    #[test]
+    fn test_transform() {
+        let url_to_title = url_titles_map(&PathBuf::from("test/url-titles.json"));
+        // println!("url_to_title: {:#?}", url_to_title);
```
caugner, line 235: Remove debug

```suggestion
```

Cargo.lock
```diff
@@ -573,6 +573,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror 2.0.12",
+ "tracing",
```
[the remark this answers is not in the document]
argl: basically logging

crates/css-syntax-types/src/lib.rs
```diff
@@ -29,14 +29,12 @@ pub mod error {
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
-#[serde(deny_unknown_fields)]
 pub struct Css {
     pub atrules: BTreeMap<String, AtRule>,
+    pub functions: BTreeMap<String, Function>,
```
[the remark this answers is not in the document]
argl: This follows a structural change in webref-css. Quoting from the [changelog](https://github.com/w3c/webref/blob/main/packages/css/CHANGELOG.md):
> Functions and types were merged in a values category in version 6, they now appear in separate functions and types categories.

crates/css-syntax-types/src/lib.rs
argl: I do not see a problem with ignoring unknown fields. The previous version was not consistent on this, so I removed those restrictions.

caugner (commented):
Second pass.

crates/css-syntax-types/src/lib.rs
argl: Yes, but I decided against it. For consistency we would have to change a lot of structs that are outside the scope of this change. I would recommend to do that in a separate PR if really needed.

caugner (approved):

PR #319, opened 2025-09-26
feat(rari-cli): print port at server startup

1ilsang:
### Description

**AS-IS**

image

**TO-BE**

image

For some reason, Rari may remain active as a zombie process.

Since it's impossible to determine which ports Rari is occupying from the logs, I want to log this information.


argl (approved):
LGTM, thanks!

LeoMcA (approved):

PR #323, opened 2025-10-08
fix(notecard): remove leading space of first paragraph in Chinese

yin1999:
### Description

remove the leading space of the first paragraph in note card

### Motivation

This PR reflects the changes made in [yari](https://github.com/mdn/yari/pull/10168), see: https://github.com/mdn/yari/pull/10168#pullrequestreview-1979828266

### DIFF
--- a/crates/rari-md/src/lib.rs
+++ b/crates/rari-md/src/lib.rs
@@ -203,6 +203,19 @@ mod test {
     }
 
     #[test]
+    fn note_zh_locale() -> Result<(), anyhow::Error> {
+        let out = m2h(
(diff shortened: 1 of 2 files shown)


caugner (commented):
Is the space really wrong, or should it just not be displayed? I'm asking to understand if we should fix this in Fred instead.

(I'm not sure if rari has other locale-specific tweaks like this.)

/cc @argl

caugner (approved):
LGTM, but requesting a second opinion from @argl.

argl (approved):
LGTM, thanks!

PR #327, opened 2025-10-13
fix(sync-translated-content): handle renames to sync case changes from content

### LINKED ISSUE #267: Sync translated content does not rewrite the slug for case-changes in the slug.

If a document is renamed (change case only) in en-US, an error around redirect map checking is raised. Compared to the yari-based version, the slug in the translated document is not being rewritten on initial investigation.

No errors and the slug being properly rewritten in that case

_No response_

_No response_

argl:
### Description

This fixes a bug for case changes of slugs.

### Motivation

Build failed in these cases.

### Additional details

Fixes #267

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2032,11 +2032,12 @@ checksum = "f5438dd2b2ff4c6df6e1ce22d825ed2fa93ee2922235cc45186991717f0a892d"
 
 [[package]]
 name = "nu-ansi-term"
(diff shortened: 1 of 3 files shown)


caugner (commented):
Could we add a test case for this change?

caugner (approved):
Tested successfully by running:

```
cargo run -- content sync-translated-content 
```

PR #337, opened 2025-10-17
fix(build): read popularities from data dir

### LINKED ISSUE #336: `doc.popularity` is always empty for all pages
This impacted full-text search, causing "sort by" to have no effect.

See:

- https://github.com/mdn/fred/issues/548
- https://github.com/mdn/fred/pull/615

---

Example: https://developer.mozilla.org/en-US/docs/Web/index.json

Image

caugner:
### Description

Fixes `popularities` to be read from the right path.

### Motivation

Ensures the `doc.popularity` is properly populated in built `index.json`.

Fixes #336.


LeoMcA (approved):
nice stuff: tested locally, and it works

Issue #366, opened 2025-11-10
Fix flaws panics at crates/rari-tools/src/fix/issues.rs:135:41


caugner:
STR:

1. Checkout content @ main
2. Clone translated-content-de @ [fix-flaws-panic](https://github.com/mdn/translated-content-de/tree/fix-flaws-panic)
3. Set `CONTENT_TRANSLATED_ROOT=../translated-content-de/files` in content `.env`
4. Run `yarn content fix-flaws --locale de` from content root

**Expected**: It should fix flaws in the German locale.

**Actual**: It panics, apparently parsing [this file](https://github.com/mdn/translated-content-de/blob/2cb09288c38ca4ad7e5ef69eac68ee2d2a513eb4/files/de/web/css/how_to/layout_cookbook/index.md?plain=1).

```
% yarn content fix-flaws --locale de                                                                                            
yarn run v1.22.22
$ yarn -s info:rari && env-cmd --silent cross-env CONTENT_ROOT=files rari content fix-flaws --locale de

🐥 This command is now using rari: https://github.com/mdn/rari
🐞 Please report any issues here:  https://github.com/mdn/rari/issues/new?template=bug.yml

Took:     2.903s for reading 28038 docs

thread '<unnamed>' panicked at crates/rari-tools/src/fix/issues.rs:135:41:
byte index 5306 is out of bounds of `---
title: CSS-Layout-Kochbuch
short-title: Layout cookbook
slug: Web/CSS/How_to/Layout_cookbook
l10n:
  sourceCommit: f3bf4e2bd456159093d3820253be9f266ace070a
---

Das CSS-Layout-Kochbuch zielt darauf ab, Rezepte für häufige Layout-Muster zusammenzufüh`[...]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error Command failed with exit code 1.
```

PR #373, opened 2025-11-10
fix(tools): calculate new offset/column properly

caugner:
### Description

Corrects the algorithm for calculating the offset after applying a fixable issue (flaw).

### Motivation

Previously, when applying a fix, the new offset was calculated incorrectly, causing out-of-bounds panics.

Additionally, we now warn and skip issues that cannot be applied.

Fixes https://github.com/mdn/rari/issues/366.


Josh-Cena:
Hello, any ETA for this? There are lots of redirect flaws in content waiting to be fixed.

PR #378, opened 2025-11-17
fix(fix-flaws): apply suggestions in order

caugner:
### Description

Rewrites the fix-flaws logic to apply suggestions in order, based on the actual location of the broken link, rather than in order of occurrence of the link (reference).

### Motivation

The `fix-flaws` command was failing in content on a page with multiple duplicate link references.

Fixes https://github.com/mdn/rari/issues/366.


argl (approved):
nice!

Issue #377, opened 2025-11-17
fix-flaws fails with JSON parse error


caugner:
Ran `yarn content fix-flaws` in mdn/content, and got the following error:

```
Updating web-platform-dx/web-features (data.extended.json) to 3.9.1
Updating @mdn/browser-compat-data to 7.1.22
Took:     3.127s for reading 28243 docs

thread '<unnamed>' panicked at crates/css-syntax/src/syntax.rs:41:41:
Failed to parse JSON: Error("missing field `name`", line: 1, column: 1704)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The issue is somewhere around here:

https://github.com/mdn/rari/blob/f9d14eb8d1dd238142e9f60cf266de31b66abcab/crates/css-syntax/src/syntax.rs#L33-L43

---

Re-running with `RUST_BACKTRACE=full` yields the following:

```
thread '<unnamed>' panicked at crates/css-syntax/src/syntax.rs:41:41:
Failed to parse JSON: Error("missing field `name`", line: 1, column: 1704)
stack backtrace:
   0:        0x105827678 - __mh_execute_header
   1:        0x1050e24b8 - __mh_execute_header
   2:        0x105827010 - __mh_execute_header
   3:        0x105827538 - __mh_execute_header
   4:        0x105826c18 - __mh_execute_header
   5:        0x105855608 - __mh_execute_header
   6:        0x105855578 - __mh_execute_header
   7:        0x105858268 - __mh_execute_header
   8:        0x10596f404 - __mh_execute_header
   9:        0x10596f7a4 - __mh_execute_header
  10:        0x1050f8fbc - __mh_execute_header
  11:        0x1050f82b0 - __mh_execute_header
  12:        0x1059ac7e0 - __mh_execute_header
  13:        0x10510a3e0 - __mh_execute_header
  14:        0x10510d878 - __mh_execute_header
  15:        0x1052ed928 - __mh_execute_header
  16:        0x1052e727c - __mh_execute_header
  17:        0x1052e54c4 - __mh_execute_header
  18:        0x1053322b8 - __mh_execute_header
  19:        0x1055be398 - __mh_execute_header
  20:        0x1055bbf28 - __mh_execute_header
  21:        0x1055bbaa0 - __mh_execute_header
  22:        0x1055bd110 - __mh_execute_header
  23:        0x1055bbe08 - __mh_execute_header
  24:        0x1055bd110 - __mh_execute_header
[...]

PR #380, opened 2025-11-19
fix(css-syntax): regenerate `css_syntax.json` on parse failure

caugner:
### Description

Updates `@webref/css` loading to regenerate the `css_syntax.json` if it failed to parse (indicating the format has changed since it was generated, most likely due to a rari update).

### Motivation

Avoids trouble for contributors when we release rari with https://github.com/mdn/rari/pull/375.

Fixes https://github.com/mdn/rari/issues/377.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,6 +573,7 @@ dependencies = [
  "serde",
  "serde_json",
  "thiserror 2.0.12",
+ "tracing",
  "url",
 ]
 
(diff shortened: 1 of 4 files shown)


caugner:
> > As I understand it, the problem is only manifests itself when people have webref 6 or lower installed, anything from version 7 onwards should not cause any errors, even the major version jump to 8 was problem-free from our perspective
> 
> I reproduced it basically as follows:
> 
>     1. Run `yarn content fix-flaws` in content.
> 
>     2. Then run `cargo run --release content fix-flaws` in rari.

I tried again, and can no longer reproduce.

Issue #361, opened 2025-11-04
The Reorganisation of CSS Content has had a knock on effect to the CSSxRef Macro


dletorey:
### What information was incorrect, unhelpful, or incomplete?

The [MDN Content](https://github.com/mdn/content) repo is having its CSS content reorganised for example see [Commit c52ed78](https://github.com/mdn/content/commit/c52ed787442db9d65b21f5c2874fa6bfd08a253a).

This means that the [`{{cssxref()}}`](https://github.com/mdn/rari/blob/29efb996fd19ba7e686d3028a45a358567a749f8/crates/rari-doc/src/templ/templs/links/cssxref.rs) macro is creating urls that then need redirecting and as there is `templ-redirected-link` and this means that the page in localhost is reloaded whatever the polling is set to.

### Example warning

```sh
[rari] serve:url:page: /en-US/docs/Web/CSS/Reference/Selectors/::part req=7 /en-US/docs/Web/CSS/Reference/Selectors/::part/index.json locale="en-US" slug="Web/CSS/Reference/Selectors/::part" file="/Users/dletorey/Git/Clients/Mozilla-dletorey/content/files/en-us/web/css/reference/selectors/_doublecolon_part/index.md"
[rari]  WARN templ{templ="cssxref" line=108 col=2 end_line=108 end_col=34}: rari_doc::templ::api: source="templ-redirected-link" ic=0 url="/en-US/docs/Web/CSS/:state" href="/en-US/docs/Web/CSS/Reference/Selectors/:state"
```

### What did you expect to see?

I would expect the slugs to be built based upon the new pattern

PR #382, opened 2025-11-19
feat(cssxref): migrate to new CSS url structure

caugner:
### Description

Updates the `cssxref` macro to use the new CSS url structure.

### Motivation

Resolves flaws caused by the CSS reorg.

### Additional details

Verified by running a build diff:

```sh
git checkout main
BUILD_OUT_ROOT=/tmp/rari/main/ cargo run --release build
git checkout 361-cssxref-reorg
BUILD_OUT_ROOT=/tmp/rari/361-cssxref-reorg/ cargo run --release build
cargo run -p diff-test diff --html --value --out diff.html --verbose /tmp/rari/main /tmp/rari/361-cssxref-reorg
```

The resulting `diff.html` only showed one difference:

image

Also verified by running `cargo run --release serve` once on rari main, and then on this branch, comparing flaws in http://localhost:8083/en-US/docs/Glossary/LTR/index.json:

#### Before

```console
% curl -s http://localhost:8083/en-US/docs/Glossary/LTR/index.json | jq '.doc.flaws'
{
  "macros": [
    {
      "type": "macros",
      "id": 0,
      "explanation": "Macro produces link /en-US/docs/Web/CSS/:dir which is a redirect",
      "suggestion": null,
      "fixable": true,
      "fixed": false,
      "line": 24,
      "column": 4,
      "endLine": 24,
      "endColumn": 23,
      "sourceContext": "- [HTML global attributes](/en-US/docs/Web/HTML/Reference/Global_attributes)\n  - [dir](/en-US/docs/Web/HTML/Reference/Global_attributes/dir)\n  - [lang](/en-US/docs/Web/HTML/Reference/Global_attributes/lang)\n- [CSS](/en-US/docs/Web/CSS)\n---^\n  - {{cssxref(\":dir\")}}\n  - {{cssxref(\"direction\")}}\n  - {{cssxref(\"unicode-bidi\")}}\n",
      "filepath": "/path/to/mdn/content/files/en-us/glossary/ltr/index.md",
      "name": "TemplRedirectedLink",
      "macroName": "cssxref",
      "href": "/en-US/docs/Web/CSS/:dir"
    },
    {
      "type": "macros",
      "id": 1,
      "explanation": "Macro produces link /en-US/docs/Web/CSS/direction which is a redirect",
      "suggestion": null,
      "fixable": true,
      "fixed": false,
      "line": 25,
      "column": 4,
      "endLine": 25,
      "endColumn": 28,
      "sourceContext": "  - [dir](/en-US/docs/Web/HTML/Reference/Global_attributes/dir)\n  - [lang](/en-US/docs/Web/HTML/Reference/Global_attributes/lang)\n- [CSS](/en-US/docs/Web/CSS)\n  - {{cssxref(\":dir\")}}\n---^\n  - {{cssxref(\"direction\")}}\n  - {{cssxref(\"unicode-bidi\")}}\n  - {{cssxref(\"writing-mode\")}}\n",
      "filepath": "/path/to/mdn/content/files/en-us/glossary/ltr/index.md",
      "name": "TemplRedirectedLink",
      "macroName": "cssxref",
      "href": "/en-US/docs/Web/CSS/direction"
    },
    {
      "type": "macros",
      "id": 2,
      "explanation": "Macro produces link /en-US/docs/Web/CSS/unicode-bidi which is a redirect",
      "suggestion": null,
      "fixable": true,
      "fixed": false,
      "line": 26,
      "column": 4,
      "endLine": 26,
      "endColumn": 31,
      "sourceContext": "  - [lang](/en-US/docs/Web/HTML/Reference/Global_attributes/lang)\n- [CSS](/en-US/docs/Web/CSS)\n  - {{cssxref(\":dir\")}}\n  - {{cssxref(\"direction\")}}\n---^\n  - {{cssxref(\"unicode-bidi\")}}\n  - {{cssxref(\"writing-mode\")}}\n",
      "filepath": "/path/to/mdn/content/files/en-us/glossary/ltr/index.md",
      "name": "TemplRedirectedLink",
      "macroName": "cssxref",
      "href": "/en-US/docs/Web/CSS/unicode-bidi"
    },
    {
      "type": "macros",
      "id": 3,
      "explanation": "Macro produces link /en-US/docs/Web/CSS/writing-mode which is a redirect",
      "suggestion": null,
      "fixable": true,
      "fixed": false,
      "line": 27,
      "column": 4,
      "endLine": 27,
      "endColumn": 31,
      "sourceContext": "- [CSS](/en-US/docs/Web/CSS)\n  - {{cssxref(\":dir\")}}\n  - {{cssxref(\"direction\")}}\n  - {{cssxref(\"unicode-bidi\")}}\n---^\n  - {{cssxref(\"writing-mode\")}}\n",
      "filepath": "/path/to/mdn/content/files/en-us/glossary/ltr/index.md",
      "name": "TemplRedirectedLink",
      "macroName": "cssxref",
      "href": "/en-US/docs/Web/CSS/writing-mode"
    }
  ]
}
```

#### After

```console
% curl -s http://localhost:8083/en-US/docs/Glossary/LTR/index.json | jq '.doc.flaws'
{}
```

Fixes https://github.com/mdn/rari/issues/361.


crates/rari-doc/src/templ/templs/links/cssxref.rs
```diff
@@ -67,11 +84,38 @@ pub fn cssxref_internal(
...
+        let url = format!("{}{}", &base_url, &url_path);
+        if RariApi::get_page_nowarn(&url).is_ok() {
+            url_path
+        } else {
+            // Fallback to Values
+            format!("Reference/Values/{slug}")
+        }
+    };
+
+    if RariApi::get_page_nowarn(&format!("{}{}", &base_url, &url_path)).is_err() {
```
LeoMcA: I don't think we should be silently falling back to redirect resolution: it's just hiding the logic of this within a massive map, while pretending the resolution is deterministic.

For instance, the only reason `CSSxRef("<url>")` resolves to [Web/CSS/Reference/Values/url_value](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/url_value) is because [Web/CSS/url](https://developer.mozilla.org/en-US/docs/Web/CSS/url) redirects there - this feels bad, and it doesn't look like anything will warn us of that.

At the very least we should get some kind of warning so we can add cases like these to the "special case mappings" `match` above, or simply move the content.

LeoMcA (approved):
With fresh eyes: this looks good, and significantly reduces the number of flaws, nice!

We can improve the special case map in future PRs

PR #384, opened 2025-11-19
ci(workflows): add build-content-and-diff

caugner:
### Description

Adds new `build-content-and-diff` workflow that runs on pull requests, and compares the build output of the PR against main.

### Motivation

Make it easy to identify the impact of changes on the build output.


LeoMcA (approved):
Very nice!

Issue #393, opened 2025-11-25
"Invalid frontmatter" errors are missing context


caugner:
See: https://github.com/mdn/translated-content/actions/runs/19654018862/job/56286826237#step:7:70

```
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
ERROR rari_doc::reader: Invalid frontmatter: could not find expected ':' at line 5 column 1, while scanning a simple key at line 4 column 1
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
Error: Fatal error reading docs
```

PR #399, opened 2025-11-27
fix(doc): add tracing context to parallel read

caugner:
### Description

Adds missing tracing context to parallel read.

### Motivation

Make the error message actionable.

#### Before

```
Building everything 🛠️
ERROR rari_doc::reader: Invalid frontmatter: duplicate field `title`
Error: Fatal error reading docs
```

#### After

```
Building everything 🛠️
ERROR page{file="/path/to/mdn/content/files/en-us/web/index.md"}: rari_doc::reader: Invalid frontmatter: duplicate field `title`
Error: Fatal error reading docs
```

Fixes https://github.com/mdn/rari/issues/393.

### DIFF
--- a/crates/rari-types/src/settings.rs
+++ b/crates/rari-types/src/settings.rs
@@ -72,7 +72,6 @@ pub struct Settings {
     pub cache_content: bool,
     pub base_url: String,
     pub live_samples_base_url: String,
-    pub legacy_live_samples_base_url: String,
     pub interactive_examples_base_url: String,
     pub additional_locales_for_generics_and_spas: Vec<Locale>,
     pub reader_ignores_gitignore: bool,


crates/rari-doc/src/reader.rs
```diff
@@ -74,6 +80,7 @@ pub fn read_docs_parallel<P: 'static + Send, T: PageReader<P>>(
                             //tx.send(Err(e.into())).unwrap();
                         }
                     }
+                    drop(enter);
```
argl: `enter` is dropped here anyway, since it is the end of its defining block. This can be removed and the `enter` variable renamed to `_enter` or `_`.

caugner: Good point.

caugner:
```suggestion
```

PR #412, opened 2025-12-03
chore(diff-test): add --terminal option

caugner:
### Description

Updates the `diff-test` crate, adding a `--terminal` option to the `diff` script.

### Motivation

Show the diff in the workflow run directly, in addition to the downloadable HTML artifact.


caugner:
Closing for now.

PR #431, opened 2025-12-10
fix(cssxref): Correct `fit-content()` slug and `calc()` example anchor

dipikabh:
### Description

This PR fixes the following:

- Updates `fit_content_function` to `fit-content_function` to match the file slug in [`fit-content()`](https://github.com/mdn/content/blob/7c5cdb65b3f9b9cfe05e34a602eaf08d61195b3e/files/en-us/web/css/reference/values/fit-content_function/index.md?plain=1#L3).
- Includes the anchor fragment in the example that shows how to use the 3rd parameter of the macro.

### Motivation

The incorrect `fit_content_function` mapping will not work for `{{cssxref("fit-content()")}}`


dipikabh:
While we're here, could we also get special handling for these two items that are both in `/Web/CSS/Reference/Values`:

- `"&lt;url&gt;" | "<url>" => "url_value"` (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/url_value)
- `"url()" => "url_function"` (https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/url_function)

Issue #369, opened 2025-11-10
Remove mdn-data dependency


caugner:
Rari currently still depends on [mdn-data](https://www.npmjs.com/package/mdn-data), even though the same data should be included in webref.

PR #434, opened 2025-12-12
chore(css_ref): remove mdn-data dependency

caugner:
### Description

Updates the `css_ref` macro to no longer use `mdn-data`.

### Motivation

We want to archive `mdn-data`, and this removes one of two usages of `mdn-data` in Rari, blocking the archival.

### Additional details

- This PR: https://rari-pr434.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference#index
- Prod: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference#index

Part of https://github.com/mdn/rari/issues/369.

Extracted from https://github.com/mdn/rari/pull/390.

### DIFF
--- a/.github/workflows/build-content-and-diff.yml
+++ b/.github/workflows/build-content-and-diff.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     paths:
       - '**/*.rs'
+      - '**/Cargo.lock'
+      - '**/Cargo.toml'
+      - '.github/workflows/build-content-and-diff.yml'
 
 concurrency:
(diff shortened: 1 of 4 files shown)


LeoMcA:
Created a very hacky [inline diff view in fred](https://github.com/LeoMcA/fred/commit/378dc7031aac59904f76ccb769d933989b4ca1fc) to view these changes more easily, and I think it's a shame to lose the qualifier in parentheses after duplicate names:

image

Could we simply check for duplicate names and qualify them with the title of the parent page in parentheses?

PR #436, opened 2025-12-12
chore(macros): mark links with data-templ-link

caugner:
### Description

Mark links generated by macros with the `data-templ-link` attribute.

### Motivation

Ensure that if they are "broken", they can be identified as macro flaws.

### Additional details

Basically, these flaws cannot be fixed by writers themselves; they need to be fixed in Rari.


caugner:
Oh, that doesn't seem to be the right approach, as it just skips the broken link / redirect check:

image

PR #448, opened 2025-12-19
chore(cli): warn in GitHub Actions without GITHUB_TOKEN

caugner:
### Description

Warn if Rari runs in GitHub Actions without `GITHUB_TOKEN`.

### Motivation

Make it easy to identify workflow runs where we forgot to provide the `GITHUB_TOKEN`.

### Additional details

Tested via 9d26e582108c98c5dc953cf245821d6b9be6e080, and works as expected:

image


LeoMcA (approved):
Nice!

PR #456, opened 2025-12-19
chore(doc): ignore shared assets when handling images

caugner:
### Description

Updates image handling to ignore shared assets.

### Motivation

Prevent errors related to shared assets from showing up in content PR Review Companion comments like [this one](https://github.com/mdn/content/pull/42408#issuecomment-3656723614).

### Additional details

It would be nice if the shared-assets repository automatically maintained a JSON file with information about all assets, but since we don't have that, we should better avoid these errors for now, as they're not actionable from a contributor's point of view.


LeoMcA (approved):
> It would be nice if the shared-assets repository automatically maintained a JSON file with information about all assets

Have you filed an issue for this? An advantage of this would be we could set the width/height attributes of these images in page (as this function does) to ensure no layout shifts

PR #457, opened 2025-12-19
chore(github): use GH_TOKEN if it exists

caugner:
### Description

Updates GitHub logic to use `GH_TOKEN` if exists as an alternative to `GITHUB_TOKEN`.

Note: Like GitHub CLI (see [docs](https://cli.github.com/manual/gh_help_environment), and [code](https://github.com/cli/go-gh/blob/a0a6e8947ae2ceedb496654757886ef41ef5ac72/pkg/auth/auth.go#L69-L75)), `GH_TOKEN` takes precedence over `GITHUB_TOKEN`.

### Motivation

Avoid problems when some workflows use `GH_TOKEN` instead of `GITHUB_TOKEN`.

### DIFF
--- a/crates/rari-deps/src/client.rs
+++ b/crates/rari-deps/src/client.rs
@@ -18,7 +18,7 @@ pub fn get(url: impl AsRef<str>) -> Result<Response, DepsError> {
             req_builder = req_builder.bearer_auth(token);
         } else if env::var("GITHUB_ACTIONS").as_deref() == Ok("true") {
             eprintln!(
-                "::warning::Cannot authenticate GitHub API request. (Provide GITHUB_TOKEN to get a higher rate limit.)"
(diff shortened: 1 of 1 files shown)


crates/rari-deps/src/client.rs
```diff
@@ -13,8 +13,8 @@ pub fn get(url: impl AsRef<str>) -> Result<Response, DepsError> {
 
     // check if the URL's host is api.github.com
     if url.host_str() == Some("api.github.com") {
-        // get the GitHub token from the environment
-        if let Ok(token) = env::var("GITHUB_TOKEN") {
+        // Use GH_TOKEN or GITHUB_TOKEN if set to avoid rate limiting.
+        if let Ok(token) = std::env::var("GH_TOKEN").or_else(|_| std::env::var("GITHUB_TOKEN")) {
```
caugner: Unnecessary prefix:
```suggestion
        if let Ok(token) = env::var("GH_TOKEN").or_else(|_| env::var("GITHUB_TOKEN")) {
```

PR #518, opened 2026-02-03
fix(css-syntax,doc,tools): update CSS slugs

tristantheb:
### Description

Additional corrections to CSS page reference links.

### Motivation

Apply the same type of correction everywhere and you may be able to reduce more errors in the rari logs at the same time.

### Additional details

_none_

Related to #508

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -256,9 +256,9 @@ checksum = "793db76d6187cd04dff33004d8e6c9cc4e05cd330500379d2394209271b4aeee"
 
 [[package]]
 name = "bytes"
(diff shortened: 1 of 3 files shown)


tristantheb:
It's really hard to find the error because Windows doesn't give me the same result as GitHub directly (due to the `mv` command not existing; it's the Orphan part that's causing the error for me). xD

I was able to correct what was possible, but as far as the organization of the `<color> | color_value` functions is concerned, unfortunately it has an external link/dependency 🤔

caugner (approved):

PR #529, opened 2026-02-12
ci(rari-npm): select Node version via .nvmrc + drop obsolete npm upgrade

caugner:
#### Summary

Updates the `publish-npm` and `publish-npm-manual` workflows:

1. Select Node version via new `rari-npm/.nvmrc` file.
2. Drop obsolete `npm upgrade`.

Also fixes inaccurately named workflow steps.

#### Related issues

Part of https://github.com/mdn/fred/issues/1300.

### DIFF
--- a/rari-npm/package.json
+++ b/rari-npm/package.json
@@ -20,6 +20,7 @@
   },
   "author": "MDN Engineering Team <mdn-dev@mozilla.com>",
   "license": "MPL-2.0",
+  "packageManager": "npm@11.6.2",
   "engines": {
     "node": ">=20"
   },


.github/workflows/publish-npm.yml
```diff
@@ -25,10 +25,6 @@ jobs:
           node-version-file: "./rari-npm/package.json"
           package-manager-cache: false
 
-      # Ensure npm 11.5.1 or later for trusted publishing
-      - name: Install latest npm
-        run: npm install -g npm@latest
```
caugner: Actually, this workflow has no `actions/setup-node` step, so this is required.

We could either pin an npm version here, or (better) add an `.nvmrc` file and use `actions/setup-node` for consistency. Otherwise we likely miss updating this from time to time.

LeoMcA (approved):

PR #530, opened 2026-02-12
feat(md): support custom heading IDs

caugner:
### Description

Add support for custom heading IDs:

```md
## Heading {#custom-id}
```

Renders to

```html
<h2 id="custom-id">Heading</h2>
```

### Motivation

Allow translated content to use the same heading IDs as en-US, making locale-agnostic fragment links work across locales.

### Additional details

This was discussed in https://github.com/orgs/mdn/discussions/387.


caugner:
Closing for, was this was mainly an exploration. We can come back to it later if need be.

PR #582, opened 2026-03-17
feat(baseline): show banner when all BCD keys have baseline High

### LINKED ISSUE #1391: "Web components" page missing Baseline status

https://developer.mozilla.org/en-US/docs/Web/API/Web_components

There should be a green "Baseline: Widely Available" banner on this page.

_No response_

This page lists two `browser-compat` keys:
```yaml
browser-compat:
  - html.elements.template
  - api.ShadowRoot
```

Apropos https://github.com/mdn/fred/issues/1321#issuecomment-4022887765

> yes, we've never handled multiple bcd keys for baseline banners: the logic gets surprisingly tricky.
> 
> In this case, though, we should probably just detect that the two keys are the same feature and show that banner: I've opened [mdn/rari#554](https://github.com/mdn/rari/pull/554) which does that.

We should go a bit further and show a baseline banner when all of the page's web features have baseline status "high". (This would result in the banner disappearing if just one of the features has a baseline reversion, but reversions from "high" are extremely rare.)

dfabulich:
### Description

Show a Baseline banner when a page has multiple `browser-compat` keys referring to multiple features in `web-features`, but all of them have baseline status `high`.

### Motivation

In https://github.com/mdn/fred/issues/1391 I identified a few dozen cases where this would add a "Baseline: Widely Available" status, including:

* https://developer.mozilla.org/en-US/docs/Web/API/Web_components
* https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API
* https://developer.mozilla.org/en-US/docs/Web/API/Media_Capture_and_Streams_API

None of those pages show a Baseline banner today, but they would all show a "Baseline: Widely Available" banner when this PR is merged.

* Fixes https://github.com/mdn/fred/issues/1391
* Relates to https://github.com/mdn/rari/pull/554

### DIFF
--- a/crates/rari-doc/src/baseline.rs
+++ b/crates/rari-doc/src/baseline.rs
@@ -22,8 +22,12 @@ static WEB_FEATURES: LazyLock<Option<WebFeatures>> = LazyLock::new(|| {
 
 /// Looks up baseline support status for the given browser compatibility keys in `WEB_FEATURES`.
 ///
-/// Returns the baseline for the keys' shared web feature, or `None` if the keys belong to
-/// different features or none, or `WEB_FEATURES` is not initialized.
+/// Returns a baseline and thus shows the banner when either:
+/// - All keys belong to the same web feature, or
+/// - The keys belong to different features but every key has baseline status High
+///   (fixes pages like "Web Components" with multiple BCD keys: mdn/fred#1391).
+///
+/// Returns `None` if any key is missing, or if keys span features and not all are High.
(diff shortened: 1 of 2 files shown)


LeoMcA:
Doing some cleanup here, as mentioned in https://github.com/mdn/fred/issues/1391#issuecomment-4082364236 we can't really merge banners from different features like this currently: we'll probably have to leverage web feature groups for this, along with a different design. Let's continue the discussion in that issue.

PR #615, opened 2026-04-08
feat(md): inject `data-sourcepos` into raw HTML `<a>` tags

caugner:
### Description

Update Markdown to HTML conversion, adding tracking of `sourcepos` for raw HTML links.

### Motivation

Allow fixing flaws in HTML links automatically, currently responsible for about 4,500 issues.


crates/rari-md/src/lib.rs
```diff
@@ -261,4 +404,194 @@ mod test {
...
+        );
+    }
+
+    #[test]
+    fn test_inject_sourcepos_in_html_block_multibyte_char_column() {
+        // 'é' (U+00E9) is 2 bytes in UTF-8, so <a> sits at byte offset 2 but
+        // character offset 1 (0-based), i.e. column 2 (1-based).
+        // The HtmlBlock path currently computes byte-based columns
+        // (start_col = lt - line_start + 1 = 2 - 0 + 1 = 3), whereas Comrak's
+        // HtmlInline sourcepos uses character-based columns.
```
LeoMcA: I don't think this is true, we have a test on the comrak side which asserts it uses a byte-based count:

https://github.com/mdn/rari/blob/160c44bbeb90d69c75fc9a6b47983fe6c4c9c97f/crates/rari-md/src/lib.rs#L257-L276

We should be consistent here, so `data-sourcepos` isn't reporting bytes in one place and chars in another.

I think it should be safe to use bytes, because we're splitting on ASCII boundaries ("<", "a", etc.)

crates/rari-md/src/lib.rs
```diff
@@ -261,4 +404,194 @@ mod test {
...
+        // This test expects character-based columns (matching the inline path)
+        // and will fail until the block path is fixed to count chars, not bytes.
```
LeoMcA: let's not leave temporal comments in the tests: "will fail", "currently" a few lines above, etc.

LeoMcA (changes requested):

crates/rari-md/src/lib.rs
caugner: Good catch, addressed via ebb987e8d790a802e2ae2083e839eeee5c4d2331.

crates/rari-md/src/lib.rs
caugner: Addressed via https://github.com/mdn/rari/commit/ebb987e8d790a802e2ae2083e839eeee5c4d2331.

LeoMcA (approved):

PR #617, opened 2026-04-09
fix(doc): lowercase local image sources

caugner:
### Description

Lowercases local image sources.

### Motivation

Avoids that `` causes the following flaw:

```
Error opening /home/runner/work/dex/dex/mdn/content/files/en-us/web/api/css_painting_api/Guide/boxbg.png: No such file or directory (os error 2)
```


LeoMcA (approved):
As we discussed: ultimately I think this should transition into a flaw, rather than something we silently fix - but since we currently silently fix it in the cloud function, doing so here is preferable.

PR #618, opened 2026-04-09
fix(doc): fall back to en-US redirect for unredirected locale URLs

caugner:
### Description

Update the redirect resolution logic, falling back to en-US redirects when a page neither exists nor has a redirect in the current locale.

### Motivation

Fix broken links in translated-content pointing to the old slug, and allow fix-flaws to automatically fix those.

### Additional details

A page moved in en-US won't get a locale redirect for locales where it was never translated, breaking any references to that slug.

We now fall back to the en-US redirect to resolve such URLs.

### DIFF
--- a/crates/rari-doc/src/redirects.rs
+++ b/crates/rari-doc/src/redirects.rs
@@ -92,6 +92,26 @@ where
     Ok(io::BufReader::new(file).lines())
 }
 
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
Nice!

PR #623, opened 2026-04-13
fix(curriculum): modules not displaying on landing page

LeoMcA:
The change in https://github.com/mdn/rari/commit/32943d5675c11fbc748e539abcade0ef6e7c792a to how we constructed the curriculum path broke how detected modules: this broke the display on the homepage, and nesting in the sidebar.


caugner (approved):
LGTM. IIUC the regression was caused by stripping the `/curriculum` path from inside the curriculum repo, causing the levels to be off by one.

PR #630, opened 2026-04-16
fix(fix_img): decode percent-encoded `src` before looking up image on disk

caugner:
### Description

Updates the `fix_img` logic to percent-decode `src` attribute values before constructing the filesystem path, so images with non-ASCII filenames (e.g. `bézier.gif`, `безымянный.png`) are correctly found on disk and have their `width`/`height` attributes set.

### Motivation

Ensure that attachments with non-ASCII characters get width/height in the rendered HTML.

### Additional details

The same class of bug existed in yari, where it manifested as an explicit warning:

```
Could not find enough matches for src: b%C3%A9zier_2_big.gif, index 0 out of bounds
```

In rari the failure was silent, the file simply wasn't found, no flaw was recorded, but no dimensions were set.

A real example from translated-content:

- `files/ru/learn_web_development/core/structuring_content/basic_html_syntax/безымянный.png` was referenced as ``.

Fixes https://github.com/mdn/rari/issues/629.

### DIFF
--- a/crates/rari-doc/src/html/fix_img.rs
+++ b/crates/rari-doc/src/html/fix_img.rs
@@ -163,80 +163,11 @@ pub fn img_size(
 
 #[cfg(test)]
 mod tests {
-    use std::path::{Path, PathBuf};
-
     use lol_html::{RewriteStrSettings, element, rewrite_str};
-    use rari_types::RariEnv;
-    use rari_types::fm_types::{FeatureStatus, PageType};
-    use rari_types::locale::Locale;
     use url::Url;
 
     use super::handle_img;
-    use crate::error::DocError;
-    use crate::pages::page::PageLike;
-    use crate::pages::types::utils::FmTempl;
-
-    struct TestPage {
-        path: PathBuf,
-        url: String,
-    }
-
-    impl PageLike for TestPage {
-        fn url(&self) -> &str {
-            &self.url
-        }
-        fn slug(&self) -> &str {
-            ""
-        }
-        fn title(&self) -> &str {
-            ""
-        }
-        fn short_title(&self) -> Option<&str> {
-            None
-        }
-        fn locale(&self) -> Locale {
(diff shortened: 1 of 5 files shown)


crates/rari-doc/src/html/fix_img.rs
```diff
@@ -30,8 +31,15 @@ pub fn handle_img(
             let src = src.to_lowercase();
             let url = base_url.parse(&src)?;
 
-            // Check if the file exists in the current locale
-            let mut file = page.full_path().parent().unwrap().join(&src);
+            // Check if the file exists in the current locale.
+            // The src may be percent-encoded (comrak encodes non-ASCII characters
+            // in URLs), so decode it before constructing the filesystem path.
+            let decoded_src = percent_decode_str(&src).decode_utf8_lossy();
```
LeoMcA: May be safer to use `decode_utf8()` and throw an error when we get badly formed utf8?

https://docs.rs/percent-encoding/2.1.0/percent_encoding/struct.PercentDecode.html#method.decode_utf8

crates/rari-doc/src/html/fix_img.rs
```diff
@@ -148,3 +160,68 @@ pub fn img_size(
...
+        .unwrap()
+    }
+
+    /// Filenames with non-ASCII characters (e.g. accents, Cyrillic) are
+    /// percent-encoded by the Markdown renderer (comrak).  `handle_img` must
+    /// decode them before constructing the filesystem path, otherwise the file
+    /// is never found and no `width`/`height` attributes are set.
+    #[test]
+    fn test_accented_filename_gets_dimensions() {
+        let tmp = std::env::temp_dir().join("rari-test-accented-img");
```
LeoMcA: Best to create a random tmp dir with `tempfile::TempDir`

LeoMcA (approved):
Couple of comments:

PR #632, opened 2026-04-17
feat(homepage): source latest news from blog

caugner:
### Description

Update the homepage SPA, using the most recently published non-sponsored blog posts as latest news instead of the hard-coded list in generic-content.

### Motivation

Ensure that the list is always up to date, reducing maintenance effort.

### Additional details

See [Latest news](https://rari-pr632.review.mdn.allizom.net/en-US/#:~:text=Latest%20news) in review deployment.

### DIFF
--- a/crates/rari-doc/src/pages/types/spa.rs
+++ b/crates/rari-doc/src/pages/types/spa.rs
@@ -212,7 +212,7 @@ impl SPA {
                     )?,
                     featured_contributor: featured_contributor(self.locale)?,
                     latest_news: ItemContainer {
-                        items: latest_news(&home_page_data.latest_news)?,
+                        items: latest_news()?,
                     },
                     recent_contributions: ItemContainer {
                         items: recent_contributions()?,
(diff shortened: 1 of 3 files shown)


crates/rari-doc/src/pages/types/spa.rs
```diff
@@ -212,7 +212,7 @@ impl SPA {
                     )?,
                     featured_contributor: featured_contributor(self.locale)?,
                     latest_news: ItemContainer {
-                        items: latest_news(&home_page_data.latest_news)?,
```
LeoMcA, line 215: Shall we remove `latest_news` from the reader (and then go and remove it from generic-content) since we don't need it any more?

LeoMcA (approved):

crates/rari-doc/src/pages/types/spa.rs
caugner: Done via f4b5da97.

PR #643, opened 2026-04-21
fix(issues): handle pointy-bracket link destinations with escaped angle brackets

caugner:
### Description

Handle pointy-bracket link destinations (`](<...>)`) with escaped angle brackets when fixing issues.

- Add candidates to `build_candidates()` that wrap the escaped href and slug in `<...>`, so the match is anchored to the pointy-bracket destination form.
- Escape the replacement the same way, so the written-back markdown stays valid.
- Add the bracket cases to the `build_candidates()` test and an end-to-end test for `(</...#\<\<_(Left_shift)>)`.

### Motivation

Ensure links written as pointy-bracket destinations whose fragment contains angle brackets can be fixed.

### Additional details

This is specific to the `](<...>)` destination form: per CommonMark, `<` and `>` are the only characters that *must* be backslash-escaped inside a pointy-bracket destination. Other characters may be escaped anywhere in a destination, but that is stylistic and not required, so this PR does not attempt to handle arbitrary escapes.

A fragment like `#<<_(Left_shift)` therefore appears as `#\<\<_(Left_shift)` in the source, while the rendered href reports the unescaped form.

Resolves the following errors when running `cargo run fix-flaws --locale ru`:

```
page: Could not locate '/ru/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#<<_(Left_shift)' before offset 0 (searched region: "") locale="ru" slug="Web/JavaScript/Guide/Expressions_and_operators" file="/path/to/mdn/translated-content/files/ru/web/javascript/guide/expressions_and_operators/index.md"
page: Could not locate '/ru/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#>>_(Sign-propagating_right_shift)' before offset 0 (searched region: "") locale="ru" slug="Web/JavaScript/Guide/Expressions_and_operators" file="/path/to/mdn/translated-content/files/ru/web/javascript/guide/expressions_and_operators/index.md"
page: Could not locate '/ru/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#>>>_(Zero-fill_right_shift)' before offset 0 (searched region: "") locale="ru" slug="Web/JavaScript/Guide/Expressions_and_operators" file="/path/to/mdn/translated-content/files/ru/web/javascript/guide/expressions_and_operators/index.md"
```


crates/rari-tools/src/fix/issues.rs
```diff
@@ -192,6 +192,22 @@ fn build_candidates(href: &str, suggestion: &str) -> Vec<Candidate> {
         candidates.push(Candidate::new(slug, slug_suggestion));
     }
 
+    if href.contains('<') || href.contains('>') {
```
LeoMcA, line 197: I'm unsure about the value of hard-coding these specific characters: can't any character be be escaped like this?

And if we don't generalize it, why would we hard code this for 3 examples which can be easily fixed by hand?

caugner: Fair point. I reframed this around the pointy-bracket destination form (`](<...>)`), where CommonMark requires `<` and `>` to be escaped, so these two characters are not an arbitrary pick. The escaped candidates are now wrapped in `<...>` so they only match that form, and the comment and PR description say so explicitly.

Other characters can be escaped anywhere in a destination, but that is stylistic rather than required (and a grep of content and translated-content finds no such cases in otherwise valid links), so I left arbitrary escapes out of scope.

caugner: Good point, I have not considered that other characters may be escaped as well, and there seem to be 6 occurrences of `\_` and 1 occurence of `\)` in Markdown link destinations.

caugner:
Closing in favor of https://github.com/mdn/translated-content/pull/38393.

PR #653, opened 2026-04-29
feat(doc): collect fragment IDs into `index.json`

### LINKED ISSUE #113: [Feature] Expose anchors in index.json

A page can have several anchors, which are used both to deep-link between content, and to link from e.g. BCD data.

However, it is not straight-forward to validate that an anchor exists on a page, without crawling the page.

(Does rari validate anchors on internal links?)

It would be nice if the index.json would include a list of all anchors in the content.

This would make it easier:

1. to validate `mdn_url`s in BCD with hashes,
2. to map `en-US` anchors to translated anchors (*).

(*) Note that BCD links with fragments uses en-US hashes, but since the `mdn_url` doesn't contain a locale, users might get redirected to another locale where the anchor doesn't exist.

_No response_

_No response_

caugner:
### Description

Collects fragment IDs per page, and exposes them on the built `index.json`.

### Motivation

First step towards validating fragment links, both in MDN content, and in BCD.

### Additional details

Walk `*[id]` in document order after `add_missing_ids` and expose the result as `fragments` on `Json*Doc`. Skip live samples and sidebar.

Fixes https://github.com/mdn/rari/issues/113.

### DIFF
--- a/crates/rari-doc/src/html/modifier.rs
+++ b/crates/rari-doc/src/html/modifier.rs
@@ -238,3 +238,44 @@ pub fn add_missing_ids(html: &mut Html) -> Result<(), DocError> {
     }
     Ok(())
 }
+
+/// Collects all fragment IDs from elements with an `id` attribute, in document order.
+///
+/// Skips the `quick_links` sidebar section id, which is extracted into the page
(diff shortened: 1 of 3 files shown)


LeoMcA (approved):
Nice!

Issue #642, opened 2026-04-21
[Bug]: None of the CSS @rule directives appear in the sidebar


hamishwillee:
### What information was incorrect, unhelpful, or incomplete?

For example 
- [additive-symbols](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@counter-style/additive-symbols). 
- All the `@media` features are not there in sidebar: https://github.com/mdn/content/pull/43831#issuecomment-4265968431
- 
I think it applies to all of them, but if not, certainly "many".

### What did you expect to see?

Sidebar links below the associated at-rule or media query

### Do you have any supporting links, references, or citations?

See above.

PR #654, opened 2026-04-29
fix(sidebar): render nested subpages as details

LeoMcA:
Relates to: https://github.com/mdn/rari/issues/642

Currently pages defined in a sidebar with:
```
type: listSubPages
depth: 2
nested: true
```

Render as a nested list, but with no details elements. This results in some weird rendering:

e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta

image

https://developer.mozilla.org/en-US/docs/Glossary/Boolean

image

With this change, those render to `<details>` elements, like so:

image

and:

image

This is particularly important for the sidebar changes proposed in https://github.com/mdn/rari/issues/642:

Without this change:

image

With this change:

image


Rumyra:
lgtm @LeoMcA

caugner:
Hm, this is weird:

> Took: 17.039215321s - 0 changes in 0 of 14380 files (0.0%)

caugner (commented):
Initial pass. Is there a chance to add a unit test?

crates/rari-doc/src/helpers/subpages.rs
```diff
@@ -209,13 +205,31 @@ pub fn list_sub_pages_nested_internal(
...
+                            badges: sub_page.status(),
+                            badge_locale: locale,
+                            code,
+                            only_en_us: sub_page.locale() != locale,
+                        },
+                        true,
+                    )?;
+                    out.push_str("</summary><ol>");
+                    out.push_str(&sub_pages_out);
+                    out.push_str("</ol></details></li>");
```
caugner: nit: could this be a separate function with a self-documenting name like `write_li_with_details`?

caugner: Allowed myself to push this change: a37a3099f76c9b3416c0470d738263ba5c37688b

caugner (approved):

caugner:
@LeoMcA I pushed the refactoring, feel free to merge if you're happy with it.

caugner:
Let's get this in, just verified locally that this still works as expected.

PR #655, opened 2026-04-30
fix(doc): apply redirects to git history at load time

### LINKED ISSUE #247: Modified date of moved page (sync-translated-content) is 1970-01-01
### URL MDN

https://developer.mozilla.org/fr/docs/Web/HTTP/Reference/Status/303

### Sur quelle section/titre porte ce problème ?

Modification date of the page/commit

### Quelle information était incorrecte, inutile ou incomplète ?

Bottom div "Help improve MDN" indicates that last modification happened on Jan 01 1970
Looks like the commit/page metadata is messed up

MDN Metadata indicate the following
* Document last modified: 1970-01-01T00:00:00.000Z

### Que vous attendiez-vous à voir ?

I highly doubt it is the real modification date.

### Avez-vous des liens, références ou citations sur ce sujet ?

_No response_

### Souhaitez-vous ajouter autre chose ?

_No response_

### Métadonnées MDN

<details>
<summary>Page report details</summary>

* Folder: `fr/web/http/reference/status/303`
* MDN URL: https://developer.mozilla.org/fr/docs/Web/HTTP/Reference/Status/303
* GitHub URL: https://github.com/mdn/translated-content/blob/main/files/fr/web/http/reference/status/303/index.md
* Last commit: https://github.com/mdn/translated-content/commit/
* Document last modified: 1970-01-01T00:00:00.000Z

</details>

caugner:
### Description

Update `git_history`, applying redirects at load time so moved pages resolve to a real history entry instead of the 1970-01-01 default.

- Relocate `GIT_HISTORY` / `git_history()` from `rari-types` to a new `rari-doc::git_history` module.
- For each `(old_url, new_url)` redirect, copy the old path's history entry to the new path (preserving any existing entry at the new path).

### Motivation

Prevent the build from falling back to 1970-01-01 for translated docs that `sync-translated-content` moved locally during the build.

Fixes #247

### DIFF
--- a/crates/rari-doc/src/git_history.rs
+++ b/crates/rari-doc/src/git_history.rs
@@ -3,6 +3,12 @@
 //! Loads `_git_history.json` artifacts (produced by `rari git-history`) for the
 //! content and translated-content roots and exposes them as a single map keyed
 //! by the relative file path used elsewhere in the build pipeline.
+//!
+//! Redirects from `_redirects.txt` are applied at load time so that pages whose
+//! source file was moved (locally by `sync-translated-content`, or in a build
+//! that ran before `git-history` could pick up a `content move`) still resolve
+//! to a real history entry instead of the 1970-01-01 default. See
+//! <https://github.com/mdn/rari/issues/247>.
 use std::collections::HashMap;
 use std::fs;
 use std::path::PathBuf;
@@ -11,7 +17,23 @@ use std::sync::LazyLock;
 use rari_types::HistoryEntry;
 use rari_types::globals::{content_root, content_translated_root};
 
+use crate::redirects::REDIRECTS;
+use crate::resolve::{UrlMeta, url_meta_from};
+
 static GIT_HISTORY: LazyLock<HashMap<PathBuf, HistoryEntry>> = LazyLock::new(|| {
+    let mut map = load_history_files();
+    apply_redirects(
+        &mut map,
+        REDIRECTS.iter().map(|(k, v)| (k.as_str(), v.as_str())),
+    );
+    map
+});
+
+pub fn git_history() -> &'static HashMap<PathBuf, HistoryEntry> {
+    &GIT_HISTORY
+}
+
+fn load_history_files() -> HashMap<PathBuf, HistoryEntry> {
     let f = content_root().join("_git_history.json");
     let mut map = if let Ok(json_str) = fs::read_to_string(f) {
         serde_json::from_str(&json_str).expect("unable to parse l10n json")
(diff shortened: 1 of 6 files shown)


crates/rari-doc/src/git_history.rs
```diff
@@ -0,0 +1,180 @@
...
+});
+
+pub fn git_history() -> &'static HashMap<PathBuf, HistoryEntry> {
+    &GIT_HISTORY
+}
+
+fn load_history_files() -> HashMap<PathBuf, HistoryEntry> {
+    let f = content_root().join("_git_history.json");
+    let mut map = if let Ok(json_str) = fs::read_to_string(f) {
+        serde_json::from_str(&json_str).expect("unable to parse l10n json")
```
LeoMcA, line 39:
```suggestion
        serde_json::from_str(&json_str).expect("unable to parse _git_history.json")
```

crates/rari-doc/src/git_history.rs
```diff
@@ -0,0 +1,180 @@
...
+    } else {
+        HashMap::new()
+    };
+    if let Some(translated_root) = content_translated_root() {
+        let f = translated_root.join("_git_history.json");
+        if let Ok(json_str) = fs::read_to_string(f) {
+            let translated: HashMap<PathBuf, HistoryEntry> =
+                serde_json::from_str(&json_str).expect("unable to parse l10n json");
```
LeoMcA, line 47:
```suggestion
                serde_json::from_str(&json_str).expect("unable to parse _git_history.json");
```

crates/rari-doc/src/git_history.rs
```diff
@@ -0,0 +1,180 @@
+//! # Git History Module
+//!
+//! Loads `_git_history.json` artifacts (produced by `rari git-history`) for the
+//! content and translated-content roots and exposes them as a single map keyed
+//! by the relative file path used elsewhere in the build pipeline.
+//!
+//! Redirects from `_redirects.txt` are applied at load time so that pages whose
+//! source file was moved (locally by `sync-translated-content`, or in a build
+//! that ran before `git-history` could pick up a `content move`) still resolve
```
LeoMcA, line 9: I'm confused by this comment, as `git-history` does run before the build:

https://github.com/mdn/dex/blob/0ccd674fe7a677dff20fc115798302fc70d71558/.github/workflows/prod-build.yml#L216-L219

LeoMcA: Oh, I see, we don't commit - so these redirects don't actually get put into `_git_history.json`, even though they exist on disk.

I'd suggest we do this resolution when we build that JSON, patching it at load time seems unnecessary and a bit confusing (that the on-disk representation differs from the in memory one).

If there's some reason we can't do it at JSON-build time, I'd at least rename the function here to clarify that it resolves to something different, maybe `file_history` or similar.

I'd also remove this comment and hang one off the `git_history` function, and make it assume a little less knowledge about the internals, something like:

> Loads a map of history entries for each file, resolving redirects when files exist but aren't yet committed (as we'd expect to see during a build after `sync-translated-content` is run)

LeoMcA (changes requested):
Code is good, I just think it's in the wrong place: see my last comment

crates/rari-doc/src/git_history.rs
caugner: Updated this error to name `_git_history.json`.

crates/rari-doc/src/git_history.rs
caugner: Updated this error to name `_git_history.json` too.

crates/rari-doc/src/git_history.rs
caugner: Agreed, `git-history` runs before the build. Since `sync-translated-content` runs first, I moved redirect propagation into `rari-tools::history` so `_git_history.json` includes the redirected paths; the loader now reads that file unchanged.

PR #656, opened 2026-04-30
chore(cli): count issues by locale

caugner:
### Description

Adds a per-locale breakdown of build issues to the build summary, splits the JSON issue dump by locale via a new `--issues-per-locale` flag, and makes the `--issues` argument optional (defaulting to `BUILD_OUT_ROOT/issues.json`).

### Motivation

The single "Found X issues in Y of Z files" summary makes it hard to see how issues distribute across locales — useful when triaging which translated trees need attention. Splitting the JSON dump per locale also makes downstream tooling (e.g. per-locale dashboards, per-locale fix runs) simpler than slicing one large `issues.json`.

### Additional details

- New summary lines after the totals, in `Locale` enum order with an `unknown` bucket appended if any issue's file path doesn't yield a locale:
  ```
  Found 24506 issues in 7623 of 50748 files
    en-US: 1205 issues in 495 files
    es: 4117 issues in 916 files
    ...
    unknown: 8 issues in 1 files
  ```
- `--issues` now takes 0 or 1 args; bare `--issues` writes to `BUILD_OUT_ROOT/issues.json`.
- `--issues-per-locale` writes `BUILD_OUT_ROOT/{locale}/issues.json` for each locale that has issues. Issues with no extractable locale (empty `file` path) are skipped here.
- Locale extraction is centralized in a small `locale_from_path()` helper that walks path components and returns the first one parseable via `Locale::from_str`.

### DIFF
--- a/crates/rari-cli/main.rs
+++ b/crates/rari-cli/main.rs
@@ -29,7 +29,7 @@ use rari_doc::pages::page::Page;
 use rari_doc::pages::types::doc::Doc;
 use rari_doc::reader::read_docs_parallel;
 use rari_doc::search_index::build_search_index;
-use rari_doc::utils::TEMPL_RECORDER_SENDER;
+use rari_doc::utils::{TEMPL_RECORDER_SENDER, locale_and_typ_from_path};
 use rari_sitemap::Sitemaps;
 use rari_tools::add_redirect::add_redirect;
 use rari_tools::fix::fixer::fix_all;
@@ -553,7 +553,9 @@ fn main() -> Result<(), Error> {
                 let mut by_locale: std::collections::BTreeMap<Option<Locale>, (usize, usize)> =
                     std::collections::BTreeMap::new();
                 for entry in events.iter() {
-                    let locale = locale_from_path(entry.key());
+                    let locale = locale_and_typ_from_path(Path::new(entry.key()))
+                        .ok()
+                        .map(|(l, _)| l);
                     let stats = by_locale.entry(locale).or_default();
                     stats.0 += 1;
                     stats.1 += entry.value().len();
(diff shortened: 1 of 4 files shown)


crates/rari-cli/main.rs
```diff
@@ -791,3 +849,11 @@ fn update(version: Option<String>) -> Result<(), Error> {
     info!("\n\nrari updated to `{}`", status.version());
     Ok(())
 }
+
+fn locale_from_path(path: &str) -> Option<Locale> {
```
LeoMcA: This could return a false locale if one appears in the absolute path on disk - not sure if it would be an edge case worth fixing, if we didn't already have a util which handles this:

https://github.com/mdn/rari/blob/673c5373f644fa52a8de2b0226e8757034e8153a/crates/rari-doc/src/utils.rs#L188

LeoMcA (approved):
Nice, just one issue:

crates/rari-cli/main.rs
caugner: Yeah, I had considered implementing it a bit like that, and considered an edge case. Could have guessed that the function exists already:

- [e3d6968](https://github.com/mdn/rari/pull/656/commits/e3d696866fe6da036a31749b7530266fc80c2cf6)[e3d6968](https://github.com/mdn/rari/pull/656/commits/e3d696866fe6da036a31749b7530266fc80c2cf6) to reuse `locale_and_typ_from_path()`
- [4a86dbd](https://github.com/mdn/rari/pull/656/commits/4a86dbdcc2118175cf4d1aef95554921493d41a3)[4a86dbd](https://github.com/mdn/rari/pull/656/commits/4a86dbdcc2118175cf4d1aef95554921493d41a3) + [b2fc40b](https://github.com/mdn/rari/pull/656/commits/b2fc40ba7523e5f5c8b9b0654311fd93356b2faf)[b2fc40b](https://github.com/mdn/rari/pull/656/commits/b2fc40ba7523e5f5c8b9b0654311fd93356b2faf) to add test cases for this function

caugner:
There are some issues associated with an "unknown" locale, presumably curriculum or generic-content:

```
Found 317 issues in 47 of 494 files
  en-US: 193 issues in 27 files
  unknown: 124 issues in 20 files
```

I'll follow-up separately.

PR #657, opened 2026-04-30
fix(doc): fallback blog posts to en-US

caugner:
### Description

Update blog post URL resolution to fall back to the en-US post when a non-en-US locale is requested with `fallback = true` (default).

### Motivation

Avoids "unknown" issues for en-US blog posts referenced on the homapage in other locales.

### Additional details

Previously the `PageCategory::BlogPost` guard rejected any non-en-US lookup unconditionally, so callers using the fallback path got `PageNotFound` for every blog URL in a non-en-US locale — even though the en-US post exists and the API contract promises a fallback.

### DIFF
--- a/crates/rari-tools/src/tests/blog_fallback.rs
+++ b/crates/rari-tools/src/tests/blog_fallback.rs
@@ -8,8 +8,6 @@ use super::fixtures::blog::BlogFixtures;
 const SLUG: &str = "test-post";
 const URL: &str = "/en-US/blog/test-post/";
 
-/// Without fallback, looking up an `/en-US/blog/...` URL with a non-en-US
-/// locale must return `PageNotFound` — this guards the `!fallback` branch.
 #[test]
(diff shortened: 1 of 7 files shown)


crates/rari-tools/src/tests/blog_fallback.rs
```diff
@@ -0,0 +1,49 @@
...
+        .expect_err("non-en-US locale without fallback should fail");
+
+    match err {
+        DocError::PageNotFound(_, PageCategory::BlogPost) => {}
+        other => panic!("expected PageNotFound(_, BlogPost), got {other:?}"),
+    }
+}
+
+/// With fallback, an `/en-US/blog/...` URL should resolve even when the
+/// caller-supplied locale isn't en-US — covers the new behavior.
```
LeoMcA: Let's remove the comments like this above each test, the function contents is enough to explain what's occurring.

LeoMcA (approved):

Issue #665, opened 2026-05-05
Unsolvable flaw with formal syntax: <anything> becomes a link


pepelsbey:
### What information was incorrect, unhelpful, or incomplete?

This flaw report [I got for this PR](https://github.com/mdn/content/pull/44015) is unsolvable:

> 
> 
> <b>Preview URLs</b> (1 page)
> 
> - [`/en-US/docs/Web/CSS/Reference/Values/var`](https://pr44015.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference/Values/var)
> 
> 
> <b>Flaws</b> (2)
> 
> *Found an unexpected or unresolvable flaw? [Please report it here](https://github.com/mdn/rari/issues/new?template=bug.yml).*
> 
> URL: [`/en-US/docs/Web/CSS/Reference/Values/var`](https://pr44015.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference/Values/var)
> Title: `var() CSS function`
> Flaw count: 2
> 
> - **broken_links**:
>   - `Link /en-US/docs/Web/CSS/Reference/Values/custom-property-name doesn't resolve`
>   - `Link /en-US/docs/Web/CSS/Reference/Values/declaration-value doesn't resolve`
> 
> *(comment last updated: 2026-05-05 14:41:50)*

<img width="806" height="278" alt="Image" src="https://github.com/user-attachments/assets/2d18fb34-2d45-45c1-86ff-70f83d7bd621" />

`<anything>` is auto-linked to non-existing pages.

```html
<a
  class="page-not-created"
  data-href="/en-US/docs/Web/CSS/Reference/Values/custom-property-name"
  title="The documentation about this has not yet been written; please consider contributing!"
  data-flaw="5"
>
  <span class="token property">
    &lt;custom-property-name&gt;
  </span>
</a>
```

### What did you expect to see?

No flaw

caugner:
We should probably not emit issues for broken links generated in the Formal syntax macro.

I could imagine though that the link itself can be useful for authors to identify that this _would_ be linked if the page existed.

@pepelsbey Should we silently not link at all, or still indicate the possible link (squiggly underline) just without emitting an issue?

pepelsbey:
Good question! I'll come back with the answer on Monday.

caugner:
Opened https://github.com/mdn/rari/pull/670 for now already to classify these broken links as `templ-broken-link`.

PR #670, opened 2026-05-07
fix(cssinfo,csssyntax): report links as `templ-*-link`

caugner:
### Description

Report broken/ill-cased/redirecting links in "Formal syntax" and "Formal definition" as `templ-broken-link`/`templ-ill-cased-link`/`templ-redirected-link` instead of `broken-link`/`ill-cased-link`/`redirected-link`.

### Motivation

Consistency. Make it easy to distinguish issues with macro-generated links from "authored" links.

### Additional details

Internal `<a>` tags synthesized by "Formal syntax" and "Formal definition" templates were never validated, so links with issues only surfaced as sourcepos-less `broken-link`/`redirected-link`.

We now validate them in `cssinfo`/`csssyntax`/`csssyntaxraw` after rendering and tag them with `data-templ-link` so `fix_link` skips them. Issues are now reported as `templ-broken-link` / `templ-redirected-link` / `templ-ill-cased-link` with macro name and line number.

Related to https://github.com/mdn/rari/issues/665.

### DIFF
--- a/crates/rari-doc/src/html/links.rs
+++ b/crates/rari-doc/src/html/links.rs
@@ -1,12 +1,15 @@
 use std::borrow::Cow;
 
+use lol_html::{RewriteStrSettings, element, rewrite_str};
 use rari_md::anchor::anchorize;
 use rari_types::fm_types::FeatureStatus;
 use rari_types::locale::Locale;
 use rari_utils::concat_strs;
 
 use crate::error::DocError;
+use crate::issues::get_issue_counter;
 use crate::pages::page::{Page, PageLike};
+use crate::redirects::resolve_redirect;
 use crate::resolve::locale_from_url;
 use crate::templ::api::RariApi;
 use crate::templ::templs::badges::{write_deprecated, write_experimental, write_non_standard};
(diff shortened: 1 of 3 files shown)


LeoMcA (approved):
Code looks good, one issue - visible in the content diff - is we no longer resolve redirects before rendering, because we only do this within a `if !templ_link` block:

https://github.com/mdn/rari/blob/17a6dd0d56ff40cf9137111b9b86eddd5b77897a/crates/rari-doc/src/html/fix_link.rs#L150

https://github.com/mdn/rari/blob/17a6dd0d56ff40cf9137111b9b86eddd5b77897a/crates/rari-doc/src/html/fix_link.rs#L234-L238

I believe that `remove_attribute` is already unconditionally applied above, so perhaps the redirect resolution should too?

Not a blocker - I can also we why we'd want to not resolve so we actually go and fix the data here - just wanted to flag it.

caugner:
> one issue - visible in the content diff - is we no longer resolve redirects before rendering

Good catch! We should report and resolve redirects, fixed via c4c4e023.

Resolving redirects without reporting is risky, as we might miss important issues that haunt us later.

caugner:
> Took: 18.704471469s - 0 changes in 0 of 14409 files (0.0%)

✅

PR #672, opened 2026-05-11
feat(domxref): build Web/API index to resolve links

caugner:
### Description

Updates the `domxref` macro to resolve API names via an Web/API index (rather than just composing `Web/API/{arg}`):

- `Web/API/Window/foo` can be referenced via `{{domxref("Window/foo")}}` or `{{domxref("foo")}}`
- `Web/API/BackgroundFetchManager/fetch` can be referenced via `{{domxref("BackgroundFetchManager/fetch")}}`
- `Web/API/Background_Synchronization_API/Reference/SyncEvent` can be referenced via `{{domxref("SyncEvent")}}`

Note: The argument of the `domxref` macro continues to be normalized:

- Space is replaced by underline (`" "` → `"_"`)
- Parentheses are trimmed (`"()"` → `""`)
- Prototype infix is removed (`".prototype."` → `"."`)
- Dots are replaced by slashes (`"."` → `"/"`)

### Motivation

Support shortcuts like `{{domxref("structuredClone")}}` avoiding `templ-redirected-link` issues.

### Additional details

Build comparison yielded:

| Source | main | branch | Δ |
|---|---:|---:|---:|
| `domxref` + `templ-redirected-link` | 6,531 | 5,697 | **−834** |
| `domxref` + `templ-broken-link` | 3,645 | 3,634 | **−11** |
| Other sources | ~12,856 | ~12,861 | ±5 |
| **Total** | **~23,035** | **~22,192** | **−843** |

~13% of `domxref` redirected/broken issues cleared; other templates unchanged.

### Related issues and pull requests

Part of https://github.com/mdn/rari/issues/683.

PR #682, opened 2026-05-11
feat(cssxref): build Web/CSS/Reference index to resolve links

caugner:
### Description

Updates the `cssxref` macro to resolve CSS feature names via a `Web/CSS` index (rather than constructing the URL from prefix/suffix heuristics and probing each candidate path with `RariApi::get_page_nowarn`):

- `Web/CSS/Reference/Properties/color` is referenced via `{{cssxref("color")}}`
- `Web/CSS/Reference/Values/color_value` is referenced via `{{cssxref("<color>")}}` or `{{cssxref("color_value")}}`
- `Web/CSS/Reference/Values/url_value` is reachable via `{{cssxref("url")}}` (previously fell back to bare `Web/CSS/url`); `url_function` via `{{cssxref("url()")}}`
- `Web/CSS/Reference/Selectors/:hover` via `{{cssxref(":hover")}}`
- `Web/CSS/Reference/At-rules/@media/color` via `{{cssxref("@media/color")}}`

The input syntax narrows the category looked up:

| Input form | Category |
|---|---|
| `<…>` / `&lt;…&gt;` | `Values` |
| `…()` | `Values` |
| `:…` | `Selectors` |
| `@…` | `At-rules` |
| bare | `Properties`, then `Values` |

Values pages whose canonical name ends in `_value` or `_function` are also indexed under the suffix-less name, so a normalized slug like `color` resolves to `Reference/Values/color_value`.

Note: existing input normalization is preserved:

- `<…>` / `&lt;…&gt;` are stripped
- Trailing `()` is stripped
- Special-case mappings (`<color>` → `color_value`, `:host()` → `:host_function`, …) still apply
- Fragments embedded in `name` (e.g. `font-variant-alternates#stylistic`) are split off before the index lookup and re-appended to the URL — previously `RariApi::get_page_nowarn` tolerated them implicitly
- When both `_value` and `_function` aliases collide on the same bare key (e.g. `Values/url`), the tie-break prefers `_value` — `()` syntax already routes functions directly to `Values/{name}_function`

### Motivation

Replace the two per-call page-existence probes used for **URL construction** inside `cssxref` (Properties-vs-Values branch + final legacy fallback) with a single in-memory index lookup, and clear `templ-redirected-link` issues caused by the old bare-slug fallback. (One `RariApi::get_page_nowarn` call remains for display-name decoration — deciding whether to wrap the text in `<…>` or append `()` based on the resolved page's type — and is out of scope here.)

### Additional details

Build comparison yielded:

| Source | main | branch | Δ |
|---|---:|---:|---:|
| `cssxref` + `templ-redirected-link` | 18 | 0 | **−18** |
| Other `templ-redirected-link` | 175 | 175 | 0 |
| `templ-broken-link` | 388 | 388 | 0 |
| `broken-link` | 575 | 575 | 0 |
| `redirected-link` | 31 | 31 | 0 |
| **Total** | **1,196** | **1,178** | **−18** |

All eliminated issues were cases where the old fallback emitted `/Web/CSS/{slug}` (which then redirected): 12× `url`, 4× `color`, 2× `position`.

Part of https://github.com/mdn/rari/issues/683. Follows the same pattern as #672 (`domxref` Web/API index).


caugner:
> Took: 17.460237301s - 4 changes in 4 of 14400 files (0.0%)

🟢 These four changes are actually improvements, and correspond to `{{cssxref("&lt;color>")}}` and `{{cssxref("&lt;position>")}}` in content (encoded `<`, literal `>`):

image

image

image

image

caugner:
> Took: 17.5240251s - 5 changes in 5 of 14400 files (0.0%)

✅ Another improvement landed, with the generalization of `:…()`: `{{cssxref(":heading()")}}` now resolves to `Web/CSS/Reference/Selectors/:heading_function` instead of `Web/CSS/Reference/Selectors/:heading`.

image

caugner:
After the latest changes, still:

> Took: 17.366998591s - 5 changes in 5 of 14400 files (0.0%)

crates/rari-doc/src/templ/templs/links/cssxref.rs
```diff
@@ -58,64 +68,75 @@ pub fn cssxref_internal(
...
+    };
+    if embedded_anchor.is_some() && anchor.is_some() {
+        let ic = get_issue_counter();
+        tracing::warn!(
+            source = "templ-invalid-arg",
+            ic = ic,
+            arg = "anchor",
+            "cssxref: `anchor` argument ignored because `name` already contains a fragment"
+        );
+    }
```
LeoMcA, line 88: Perhaps we should have a flaw for *all* embedded anchors so we go and normalize them all to use the anchor arg?

crates/rari-doc/src/templ/css_feature_index.rs
```diff
@@ -0,0 +1,351 @@
...
+}
+
+static CSS_FEATURE_INDEX: LazyLock<HashMap<String, Vec<String>>> = LazyLock::new(build_index);
+
+fn build_index() -> HashMap<String, Vec<String>> {
+    // Slugs are locale-invariant in MDN content, so it's enough to walk the
+    // en-US tree — the resulting paths are reused unchanged across locales
+    // when constructing URLs.
+    let pages = get_sub_pages("/en-US/docs/Web/CSS", None, SubPagesSorter::Slug)
+        .expect("failed to build cssxref CSS feature index from /en-US/docs/Web/CSS");
```
LeoMcA: Ah, we do panic here, so we should do the same in domxref, or not do so here. Ref: https://github.com/mdn/rari/pull/672#discussion_r3243012093

crates/rari-doc/src/templ/css_feature_index.rs
```diff
@@ -0,0 +1,351 @@
...
+///   the leaf alias.
+///
+/// The full sub-path (e.g. `Reference/Properties/color`) is intentionally
+/// not indexed: no content uses `{{cssxref("Reference/…")}}` in practice.
+fn index_one(map: &mut HashMap<String, Vec<String>>, sub_slug: &str) {
+    if sub_slug.is_empty() {
+        return;
+    }
+
+    let Some(after_ref) = sub_slug.strip_prefix(REFERENCE_PREFIX) else {
```
LeoMcA: `strip_prefix` will return `None` if the string doesn't start with the prefix: https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix. I suppose this is intentional, as we don't want to index non-reference pages, but up on [L45](https://github.com/mdn/rari/pull/682/changes#diff-422d734f2c3aab6a9c40545835daa9a773b7bace4b282bf200755eab3a94a5a2R45) we get all CSS pages, including non-reference, and strip the `Web/CSS` slug off: we can collapse that into just getting reference pages and stripping the slug off in one go

crates/rari-doc/src/templ/css_feature_index.rs
```diff
@@ -0,0 +1,351 @@
...
+        return;
+    };
+    map.entry(after_ref.to_lowercase())
+        .or_default()
+        .push(sub_slug.to_string());
+
+    let Some((category, name)) = after_ref.split_once('/') else {
+        return;
+    };
```
LeoMcA, line 113: Should the logic be a bit stricter here, and only continue if `category` exists in the `CssRefCategory` enum?

crates/rari-doc/src/templ/css_feature_index.rs
```diff
@@ -0,0 +1,351 @@
...
+    key.push_str(cat);
+    key.push('/');
+    key.push_str(slug);
+    key.make_ascii_lowercase();
+    let candidates = map.get(&key)?;
+    candidates
+        .iter()
+        .min_by_key(|v| {
+            // Every indexed canonical starts with `Reference/` (see `index_one`).
+            let after_ref = &v[REFERENCE_PREFIX.len()..];
```
LeoMcA: see: https://github.com/mdn/rari/pull/682/changes#r3258351533

LeoMcA (changes requested):
Let's standardise the panic (or not) behaviour with domxref, and fix the double slug strip which should collapse a bit of this logic

crates/rari-doc/src/templ/templs/links/cssxref.rs
caugner: My take here is that we should rather get rid of the anchor argument long-term, because it is more natural to combine the slug with the anchor.

crates/rari-doc/src/templ/css_feature_index.rs
caugner: Yes, I agree, we should panic.

crates/rari-doc/src/templ/css_feature_index.rs
caugner: Fixed via 3d5fb25e.

crates/rari-doc/src/templ/css_feature_index.rs
caugner: Fixed via 04a2e7b4.

crates/rari-doc/src/templ/css_feature_index.rs
caugner: Fixed via 3d5fb25e.

crates/rari-doc/src/templ/templs/links/cssxref.rs
```diff
@@ -27,21 +29,29 @@ use crate::templ::api::RariApi;
...
-/// The macro generates URLs based on the new CSS reference organization:
-/// - Data types (starting with `<` or `&lt;`): `/Web/CSS/Reference/Values/{slug}`
-/// - Pseudo-classes/elements (starting with `:`): `/Web/CSS/Reference/Selectors/{slug}`
-/// - At-rules (starting with `@`): `/Web/CSS/Reference/At-rules/{slug}`
-/// - Functions (ending with `()`): `/Web/CSS/Reference/Values/{slug}`
-/// - Properties: `/Web/CSS/Reference/Properties/{slug}` (checked first)
-/// - Other values: `/Web/CSS/Reference/Values/{slug}` (fallback)
-/// - If page not found in new structure: fallback to `/Web/CSS/{slug}`
+/// The macro resolves the (normalized) name against an index of all
+/// `Web/CSS/*` pages (see [`crate::templ::css_feature_index`]). The input
```
LeoMcA:
```suggestion
/// `Web/CSS/Reference/*` pages (see [`crate::templ::css_feature_index`]). The input
```

LeoMcA (approved):
Lgtm!

caugner:
> Took: 17.978137277s - 211 changes in 29 of 14427 files (0.2%)

As far as I can see, all changes are expected, and correspond to:
1. resolved flaws (yay!),
2. actually fixed links:
  - `<color>` previously linked to `Properties/color`, and now links to `Values/color_value` (same for `<position>`)
  - `:heading()` previously linked to `Selectors/:heading`, and now links to `Selectors/:heading()`
3. changes some flaws to be more precise:
  - `<url>` now points to `Values/url` 

ezgif-63c61a7db83bb13b

PR #684, opened 2026-05-12
feat(templ): support `_<xref>()` to mark intentionally missing pages

caugner:
### Description

Adds a leading-underscore syntax (`{{_cssxref("foo")}}`) for xref macros to mark links whose target page is intentionally not-yet-created. Missing pages flagged this way no longer emit `templ-broken-link`; if the target later starts resolving, a new `templ-expect-missing-exists` flaw fires and is auto-fixable via `content fix-flaws`.

### Motivation

Authors today have no way to distinguish "this xref is broken" from "this xref points at a page I'm about to add". The first case is a real flaw; the second is noise. Marking the latter explicitly keeps the broken-link signal clean and gives us a way to detect when the placeholder is no longer needed.

### Additional details

- Grammar change in mdn/tree-sitter-mdn#103 (temporary path-dep workaround in 145b5128, to be reverted on release).
- Opt-in per template via `#[rari_f(accepts_expect_missing = true)]`; only `cssxref` is opted in here — follow-ups can extend to the other xref-family templates (`domxref`, `jsxref`, `htmlxref`, …).
- Flaw resolution lives in `RariApi::get_page_internal`, driven by a thread-local set around each template invocation (mirrors the existing `TEMPL_RECORDER` pattern).
- Verified end-to-end against `mdn/content`: existing-page invocations flag and auto-fix, missing-page invocations correctly suppress the broken-link warn.

**Depends on:** mdn/tree-sitter-mdn#103

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -51,6 +51,11 @@ jobs:
           path: mdn/rari
           persist-credentials: false
 
+      # TODO: remove once tree-sitter-mdn with expect-missing support is released
+      - name: Checkout tree-sitter-mdn PR #103
+        if: inputs.rari-binary-artifact-id == ''
+        run: git clone --depth 1 --branch expect-missing https://github.com/mdn/tree-sitter-mdn.git mdn/tree-sitter-mdn
+
       - name: Checkout (content)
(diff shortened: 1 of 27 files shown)


caugner:
The content team has expressed their preference for a list of known missing pages, so I'm closing this PR, which was just a quick prototype.

PR #685, opened 2026-05-12
feat(csssyntax): link constituent expansion headings

### LINKED ISSUE #165: Should other attributes be linked for consistency in the formal syntax?
### Summary

example on https://developer.mozilla.org/en-US/docs/Web/CSS/calc#formal_syntax with https://developer.mozilla.org/en-US/docs/Web/CSS/calc-keyword 
→ ❓Probably makes sense to link somehow, but needs discussion, yet is low priority. Wait for the rewrite to be done, then bring it up in 2025.
From @LeoMcA : I believe this is because the <calc-keyword> definition is expanded in calc(), so there's no need to link to it to explain it - whether it should still be linked for consistency with the other keywords is another question

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/calc#formal_syntax

### Reproduction steps

1. Copy the link in the browser URL bar https://developer.mozilla.org/en-US/docs/Web/CSS/calc#formal_syntax 
2. Observe the linked attributes.
3.Observe that not all attributes are links, despite existing reference pages. 

### Expected behavior

Discuss and establish whether all attributes should be linked for consistency. 

### Actual behavior

Inconsistent behavior. 
image

### Device

Desktop

### Browser

Chrome

### Browser version

Stable

### Operating system

Android

### Screenshot

image

### Anything else?

_No response_

### Validations

- [X] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [X] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [X] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
### Description

A formal-syntax box expands referenced types inline, but the expansion's heading (`<length-percentage> = `) was plain text. This links it to that type's own page.

- `SyntaxLine` gains a `heading_url`, set for constituent expansions only.
- The URL comes from the `CssRefResolver`, so an undocumented production keeps a plain heading rather than gaining a broken link.
- `CssRefKind::AtRule` is added for `@media`-style expansions.

### Motivation

Where a type is explained, there was no way to reach its reference page. Linking the heading rather than every inline occurrence keeps the box readable: one link per expansion instead of one per mention.

### Additional details

Verified over all 1256 en-US `Web/CSS` pages: 393 pages differ, **every one of them only by heading links** (1096 of 3121 headings gain one), the heading set itself is unchanged, and the issue count is unchanged at 114, so no new broken links.

Reworked from the previous revision of this PR, which predates the resolver:

- `value_heading_url` hardcoded the same five-entry slug map that #877 deletes, and built URLs unconditionally, so undocumented constituents (`<style-feature>`, `<container-condition>`, …) would have gained dead links. Both are now handled by the resolver.
- Headings resolving to the page being rendered are left unlinked, via `RefLinks::page_path`, so `<anchor-size>` on `Values/anchor-size` and `<content()>` on `Properties/content` don't link to themselves. Without this, 5 pages got a self-link.
- Constituent dedup now compares name/syntax/specs explicitly instead of relying on `SyntaxLine`'s derived `PartialEq`. Because `heading_url` differs per reference kind, a type reached as both `<x>` and `<'x'>` would otherwise be expanded twice (`<integer>` was duplicated on 23 pages).
- The resolver and page path are bundled into `RefLinks`, which keeps `render_formal_syntax` under clippy's argument limit.
- Because the existing tests render with `resolver: None`, they need no snapshot changes; the previous revision had to rewrite two large expected strings.

Not addressed here: 30 *inline* self-links remain (e.g. `<color>` on `Values/color_value`). Those predate this stack and are a separate question.

Fixes #165.
Depends on:- #877


caugner:
@Josh-Cena @Rumyra Can you please take a look at the screenshots above and let me know if that makes sense? Basically, linking to types unconditionally adds 7 links in the above example. (I think it's fine.)

caugner:
> Basically, linking to types unconditionally adds 7 links in the above example.

Instead, we could only link the heading of the expanded constituent types (note: this would need some style adjustments in fred):

image

Josh-Cena:
Does the red squiggly mean "broken"? Is the intention that this page should be eventually written (which I doubt will; a lot of these types are strictly internal and purely for spec convenience)?

caugner:
In Rari a broken link is just a link where we know that the target does not exist, so the link rendering is replaced by a placeholder element. How that element is styled is up to Fred, and currently these are displayed with squiggly underline.

Josh-Cena:
Per my understanding, there are far more types in the CSS spec than we'd like to document; most of them are only ever used once or twice as an intermediate form that composes with other stuff. So:

- If the type already has a documentation page, great, link to it no matter if it's expanded on this page or not;
- If the type has no documentation page, I think creating a link for it (no matter how it's styled) makes less sense than not linking to it, because it's more likely that a page will never exist.

caugner:
> * If the type already has a documentation page, great, link to it no matter if it's expanded on this page or not;

One important question is where it should be linked:

1. Only on the expansion.
2. Only on the 1st reference.
3. On the expansion and each reference.

Otherwise it is absolutely possible to only link where a page exists, but that is a slightly different issue as #165 where links on expanded attributes are currently missing.

Rumyra:
We had a semi related discussion in the editorial last week - everyone was agreed formal syntax can be 'over linked' at times - so...

I like only linking to pages if they exist and the option of only on the expansion

caugner:
> We had a semi related discussion in the editorial last week - everyone was agreed formal syntax can be 'over linked' at times - so...
> 
> I like only linking to pages if they exist and the option of only on the expansion

Just to be clear, in the example above, would you expect the following links to be added:

image

(Given that https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc-keyword and https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc-sum exist, but not `<calc-product>` nor `<calc-value>`.)

Edit: Or would you rather expect them to be linked where they're referenced (this is consistent with the other values being linked, but makes it less clear which ones are expanded, and which ones aren't):

image

Josh-Cena:
I would not mind the left-hand side of the definition being linked, as long as it's not to the current page.

Rumyra:
The first example is better - agree with Josh - left hand as long as not on same page

crates/css-syntax/src/syntax.rs
```diff
@@ -537,27 +580,40 @@ impl SyntaxRenderer<'_> {
...
                         let mut syntax = get_syntax(CssType::Property(&property.name), None);
                         syntax.name = format!("<{}>", syntax.name);
+                        syntax.heading_url =
+                            self.heading_url(CssRefKind::Property, &property.name);
                         Some(syntax)
                     }
                     // Node::Function(function) => Some(get_syntax(CssType::Function(&function.name))),
                     Node::AtKeyword(at_keyword) => {
-                        Some(get_syntax(CssType::AtRule(&at_keyword.name), None))
+                        let mut syntax = get_syntax(CssType::AtRule(&at_keyword.name), None);
```
LeoMcA: I think there's an issue here: `at_keyword.name` excludes the leading `@`, whereas I believe `get_syntax` requires one

LeoMcA (changes requested):

crates/css-syntax/src/syntax.rs
caugner: Good catch, though adding `@` did not cause any build changes, so I will revert corresponding changes.

caugner:
| [Before](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc#formal_syntax) | [After](https://rari-pr685.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference/Values/calc#formal_syntax) |
|--------|--------|
| image | image |

LeoMcA (approved):
Lgtm!

PR #686, opened 2026-05-13
fix(templ): accept slug for subpage macros

caugner:
### Description

Updates the `{{QuickLinksWithSubpages}}` and `{{ListSubpagesForSidebar}}` macros to accept a slug (without `/{locale}/docs/` prefix), without triggering a flaw.

### Motivation

Avoid unnecessary flaws (the slug in the context of a locale is not ambiguous).

### Additional details

Previously, this triggered flaws like this:

```
fixed legacy url: Learn/Common_questions -> /es/docs/Learn/Common_questions
```

Related to https://github.com/mdn/translated-content/pull/36029.

### DIFF
--- a/crates/rari-doc/src/templ/legacy.rs
+++ b/crates/rari-doc/src/templ/legacy.rs
@@ -1,104 +1,83 @@
-use std::borrow::Cow;
-use std::str::FromStr;
-
 use rari_types::locale::Locale;
-use tracing::warn;
 
-pub fn fix_broken_legacy_url(url: &str, locale: Locale) -> Cow<'_, str> {
-    let first_non_slash = url.find(|c| c != '/').unwrap_or(url.len());
-    let trimmed = &url[first_non_slash..];
-    let maybe_locale = &trimmed[..trimmed.find('/').unwrap_or(trimmed.len())];
-    let parsed_locale = Locale::from_str(maybe_locale);
-    let has_locale = parsed_locale.is_ok();
-    let (has_docs, maybe_without_docs) = if has_locale {
-        let without_locale = trimmed.strip_prefix(maybe_locale).unwrap_or(trimmed);
-        (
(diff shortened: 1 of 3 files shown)


crates/rari-doc/src/templ/legacy.rs
```diff
@@ -1,104 +1,83 @@
...
+    #[test]
+    fn prefix_check_is_locale_specific() {
+        // A `/fr/docs/...` URL processed under en-US is treated as a slug
+        // and gets re-prefixed.
         assert_eq!(
-            fix_broken_legacy_url("/Web", Locale::EnUs),
-            "/en-US/docs/Web"
+            normalize_url("/fr/docs/Web", Locale::EnUs),
+            "/en-US/docs/fr/docs/Web"
         );
```
LeoMcA, line 76: This feels like wrong behaviour

crates/rari-doc/src/templ/legacy.rs
```diff
@@ -1,104 +1,83 @@
...
+        return String::new();
+    }
+    let prefix = format!("/{}/docs", locale.as_url_str());
+    if input.starts_with(&prefix) {
+        input.to_string()
     } else {
-        (trimmed.starts_with("docs/"), trimmed)
-    };
+        format!("{prefix}/{}", input.trim_start_matches('/'))
+    }
```
LeoMcA, line 17: There's a number of helper functions in `crates/rari-doc/src/resolve.rs` which feel like they'd be useful here

LeoMcA (commented):
This PR feels in conflict with https://github.com/mdn/translated-content/pull/36029 - I'm not sure why we need to "fix" behaviour in rari if we can fix it in translated content, especially because these macros aren't used at all in en-US

caugner:
Closing in favor of https://github.com/mdn/rari/pull/866.

PR #696, opened 2026-05-15
feat(popularity): include redirected URLs in page popularity

caugner:
### Description

Aggregate popularity across redirects: a page's effective popularity is now its own page views plus the page views of every URL that redirects to it. Applied to both the per-page `popularity` in built JSON and the ordering of `search-index.json`.

### Motivation

When a page is moved (e.g. `Web/Foo/bar` → `Web/Reference/Foo/Bar`), pageview history attributed to the old URL in `popularities.json` was previously lost, making the moved page appear less popular than it actually is and ranking it lower in search.

### Additional details

- Adds `REDIRECTS_REVERSED` (lowercase target URL → source URLs) and a `popularity_for(url)` helper in `rari-doc/src/redirects.rs`.
- Lowercases `popularities.json` keys at load so the cross-lookup against the lowercase-keyed `REDIRECTS` map is consistent.
- `_redirects.txt` files contain no chains, so a single hop suffices.
- Real-world impact on en-US `search-index.json` is currently small: only 6 of 17.5k redirect sources appear in `popularities.json` (the upstream feed is keyed by current canonical URLs). Top-100 ordering is unchanged; the 5 pages that move up are:

| main rank | branch rank | Δ | URL |
|---:|---:|---:|---|
| 13358 | 3180 | −10178 | `/en-US/docs/Glossary/Style_sheet` |
| 14372 | 10194 | −4178 | `/en-US/docs/WebAssembly/Reference/Numeric/shr_s` |
| 14314 | 10531 | −3783 | `/en-US/docs/WebAssembly/Reference/Definitions/global` |
| 14338 | 10705 | −3633 | `/en-US/docs/WebAssembly/Reference/SIMD/load_store/load` |
| 154 | 143 | −11 | `/en-US/docs/Learn_web_development/Core/Scripting` |


crates/rari-doc/src/redirects.rs
```diff
@@ -161,6 +179,34 @@ pub fn resolve_redirect<'a>(url: impl AsRef<str>) -> Option<Cow<'a, str>> {
     }
 }
 
+/// Returns the aggregated popularity for `url`: the page's own page views plus
+/// the page views of every URL that redirects to it.
+///
+/// Returns `None` if neither the page nor any URL redirecting to it has an
+/// entry in `popularities.json`.
+pub fn popularity_for(url: &str) -> Option<f64> {
```
caugner, line 187: Maybe this should go in a different module.

LeoMcA: Agreed, the consumer is search, so perhaps search_index.rs?

crates/rari-doc/src/redirects.rs
```diff
@@ -205,4 +251,95 @@ mod tests {
...
+
+    #[test]
+    fn popularity_only_from_redirect_sources() {
+        let r = reverse(&redirects(&[("/en-US/docs/Old", "/en-US/docs/New")]));
+        let p = pops(&[("/en-US/docs/Old", 2.5)]);
+        assert_eq!(aggregate_popularity("/en-us/docs/new", &p, &r), Some(2.5));
+    }
+
+    #[test]
+    fn popularity_returns_none_when_absent() {
```
LeoMcA, line 324: For completeness, let's add a similar test like:

```
redirects = foo -> bar
pops = bar: 1, other: 2
assert aggregate(other) = 2
```

crates/rari-doc/src/redirects.rs
```diff
@@ -205,4 +251,95 @@ mod tests {
...
+        let r = reverse(&redirects(&[]));
+        let p = pops(&[("/en-US/docs/Other", 1.0)]);
+        assert_eq!(aggregate_popularity("/en-us/docs/missing", &p, &r), None);
+    }
+
+    #[test]
+    fn popularity_normalizes_casing_via_public_fn_inputs() {
+        // The public `popularity_for` lowercases its input; verify that the
+        // aggregate helper handles already-lowered input consistently across
```
LeoMcA, line 333: I don't see what this is testing: should the redirects or popularities slugs already be lowercased here?

crates/rari-types/src/globals.rs
```diff
@@ -167,11 +167,17 @@ pub static POPULARITIES: LazyLock<Popularities> = LazyLock::new(|| {
...
-    }
+    let Ok(json_str) = fs::read_to_string(f) else {
+        return Popularities::default();
+    };
+    let mut p: Popularities =
+        serde_json::from_str(&json_str).expect("unable to parse popularities json");
+    p.popularities = p
+        .popularities
+        .into_iter()
+        .map(|(k, v)| (k.to_ascii_lowercase(), v))
```
LeoMcA, line 178: nit: we use `to_lowercase` in redirects.rs, but `to_ascii_lowercase` here - shouldn't be an issue with our slugs, but let's standardise?

LeoMcA (changes requested):

crates/rari-doc/src/redirects.rs
caugner: I was thinking of a specific popularity module?

caugner: Done. Moved the inverted redirect map and `popularity_for` into `crates/rari-doc/src/popularity.rs`; built page JSON and the search index now use that module.

crates/rari-doc/src/redirects.rs
caugner: Added a test for `foo -> bar` while `other` has its own popularity entry, confirming unrelated pages keep their own score.

crates/rari-doc/src/redirects.rs
caugner: Clarified the lookup boundary: `aggregate_popularity` normalizes the URL before looking it up, and the test passes a mixed-case URL.

crates/rari-types/src/globals.rs
caugner: Changed the popularities loader to use `to_lowercase()`, matching the redirect URL normalization.

PR #697, opened 2026-05-15
feat(cli): accept positional FILES on build, deprecate -f/--files

caugner:
### Description

Updates `rari build` to accept file paths as positional arguments. Also hides the existing `-f`/`--files` flag from `--help` and emits a deprecation warning when used on its own, while rejecting it when mixed with positional `FILES`.

### Motivation

Enables natural shell composition with tools like `xargs`, e.g.:

```sh
git ls-files -m '*.md' | xargs cargo run -- build
```

Previously each path had to be prefixed with `-f`, which doesn't compose with `xargs` without extra plumbing.

### Additional details

- Positional `FILES` has identical semantics to `-f` today: triggers `Cache::Dynamic` and skips the full content traversal.
- `--file-list` is unchanged.
- `-f`/`--files` will be removed in a future release once downstream scripts have migrated.
- Invalid positional paths now report all offending entries at once (previously only the first failure surfaced).

### DIFF
--- a/crates/rari-cli/main.rs
+++ b/crates/rari-cli/main.rs
@@ -198,8 +198,16 @@ struct ServeArgs {
 
 #[derive(Args)]
 struct BuildArgs {
-    #[arg(short, long, help = "Build only content <FILES>")]
+    #[arg(value_name = "FILES", help = "Build only these content files")]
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
Nice

PR #698, opened 2026-05-15
feat(cli): add `--locale` build option

caugner:
### Description

Update `rari build` to accept a repeatable/comma-separated `--locale` flag:

- Scopes the doc-read step and SPA emission to the requested locales, even when `content_translated_root` is configured.
- en-US is always included so cross-locale links, sidebars, and templates still resolve.
- Unsupported `--locale` values (locales that aren't actively translated) are rejected with an error listing the supported set.
- Locale selection is now expressed via a `LocaleFilter::All` / `LocaleFilter::Only(&[Locale])` enum instead of an ambiguous `Option<&[Locale]>`, making call-site intent explicit.
- The same filter and `--locale` validation is wired into `content fix-flaws --locale`: it no longer reads (then discards) every non-requested translated locale, and invalid or unsupported locale values are now rejected with the same error as `build`.

### Motivation

Make it easy to build a single locale (e.g. en-US) without having to uncomment `content_translated_root`.

### Additional details

Wall-clock comparison on a full content checkout (14,396 docs):

| Build | SPAs | Wall |
|---|---|---|
| `build`, translated root unset | 14 | 10.06 s |
| `build --locale en-US` (translated root configured) | 14 | 10.50 s |
| `build` (translated root configured, all locales) | 118 | significantly higher |

Example:

```console
% cargo run --quiet --release build --locale fr
Building locale(s): en-US, fr 🛠️
Took:  974.447ms for reading 21724 docs
Took:  116.237ms to build spas (27)
Took:    18.778s to build content docs (21724)
Took:   12.752ms to build search index
Found 3951 issues in 1392 of 21751 files
  en-US: 1111 issues in 445 files
  fr: 2840 issues in 947 files
```

`content fix-flaws --locale fr` benefits too — read phase ~3× faster (50,984 → 21,725 pages, 2.5–2.7 s → 0.86–0.99 s), shaving ~2 s off total wall time.


crates/rari-cli/main.rs
```diff
@@ -263,6 +263,12 @@ struct BuildArgs {
         help = "Noop flag to legacy compatibility (has no effect on build)"
     )]
     noop: bool,
+    #[arg(
+        long,
+        value_delimiter = ',',
+        help = "Only build the given locale(s). Repeatable or comma-separated. en-US is always included. Non-en-US values require `content_translated_root`."
```
caugner:
```suggestion
        help = "Only build the given locale(s). Repeatable or comma-separated. en-US is always included. Other locales require `content_translated_root`."
```

crates/rari-cli/main.rs
```diff
@@ -625,7 +674,7 @@ fn main() -> Result<(), Error> {
             settings.data_issues = true;
             settings.blog_unpublished = true;
             let _ = SETTINGS.set(settings);
-            read_and_cache_doc_pages()?;
+            read_and_cache_doc_pages(None)?;
```
caugner: This reads weirdly, I wonder if this option should be an enum `LocaleFilter::some()` / `LocaleFilter::unrestricted()`.

caugner: 676ccfa93db029dabda74d13ce30f8b537c7dd8e

crates/rari-doc/src/pages/types/spa.rs
```diff
@@ -517,4 +526,23 @@ mod test {
     fn print() {
         println!("{}", serde_json::to_string(&*BASIC_SPAS).unwrap())
     }
+
+    #[test]
+    fn all_without_filter_returns_pairs() {
+        let pairs = SPA::all(LocaleFilter::All);
+        assert!(!pairs.is_empty());
+        assert!(pairs.iter().any(|(_, locale)| *locale == Locale::EnUs));
```
LeoMcA: This isn't meaningfully different from the test below: we're not actually testing that other locales appear

crates/rari-doc/src/pages/types/spa.rs
```diff
@@ -517,4 +526,23 @@ mod test {
...
+    }
+
+    #[test]
+    fn all_with_en_us_filter_yields_only_en_us() {
+        let pairs = SPA::all(LocaleFilter::Only(&[Locale::EnUs]));
```
LeoMcA, line 545: We should test another locale to ensure that is added as expected

crates/rari-cli/main.rs
```diff
@@ -689,7 +747,12 @@ fn main() -> Result<(), Error> {
                 // Collect content pages
                 if fix_content {
                     let start = std::time::Instant::now();
-                    let docs = read_and_cache_doc_pages()?;
+                    let only = args.locale.map(|l| [l]);
+                    let locale_filter = match only.as_ref() {
+                        Some(arr) => LocaleFilter::Only(arr.as_slice()),
+                        None => LocaleFilter::All,
+                    };
+                    let docs = read_and_cache_doc_pages(locale_filter)?;
```
LeoMcA, line 915: This skips the validation which the build command does above, like warning content_translated_root needs to be defined, the locales exist, etc.

Could we extract that into a helper and use it in both places?

LeoMcA (changes requested):

crates/rari-doc/src/pages/types/spa.rs
caugner: Fixed via 46df2836b60c6d6db91363f284d5da8bbfae5a88.

crates/rari-doc/src/pages/types/spa.rs
caugner: Fixed via 46df2836b60c6d6db91363f284d5da8bbfae5a88.

crates/rari-cli/main.rs
caugner: Resolved via b00d9d27a80f7f45692156006697902aa9635b46.

crates/rari-doc/src/pages/types/spa.rs
```diff
@@ -541,6 +547,15 @@ mod test {
         assert!(pairs.iter().all(|(_, locale)| *locale == Locale::EnUs));
     }
 
+    #[test]
+    fn all_with_fr_filter_yields_only_fr_locales() {
+        // Only fr locales must appear; en-US (or any other locale) must not.
+        // When no translated root is configured the result is empty, which
+        // satisfies the assertion trivially.
```
LeoMcA: I think these tests aren't really making sense because we don't have a translated root defined, so only en-us can ever exist: can we mock this somehow in the tests?

caugner: My understanding is that we do have a translated root defined in tests. 41633010 clarifies this now.

LeoMcA: Ok nice!

LeoMcA (approved):

PR #700, opened 2026-05-15
fix(issues): fix corrupted positions in nested template macros

caugner:
### Description

Fix `populate_issue_from_scope` to take `(line, col, end_line, end_col)` from a single scope frame instead of merging fields independently across frames.

### Motivation

Prevent inverted/nonsensical column ranges (e.g. `col=23, end_col=11`) on issues raised inside nested templ renders. When a `templ` function recursively renders an inner fragment via `render_and_decode_ref` (e.g. `cssxref` inside `cssinfo`), the inner `render` call creates a nested `templ` span whose `line`/`col`/`end_col` refer to positions inside a synthetic intermediate string, not the markdown source. The previous per-field merge — using `0` as the "unset" sentinel — mixed the inner span's `col=23` with the outer span's `end_col=11`, producing the inverted range.

### Additional details

`scope()` iterates innermost-first, so taking positions from the first frame with a real `line` set means innermost-with-real-positions wins, and nested synthetic-string spans (which carry `line=0`) no longer corrupt the outer positions.

A regression test in `issues::tests` verifies the scenario directly: an inner synthetic span (`line=0, col=23`) nested inside a real outer span (`line=5, col=0, end_col=11`) — the test asserts that `col=0` (outer) wins over `col=23` (inner), and that all four position fields are taken from the same frame.


crates/rari-doc/src/issues.rs
```diff
@@ -171,17 +171,18 @@ where
...
-            }
-            if entries.line != 0 {
+            // Position fields form an atomic group: a nested `templ` span (e.g. cssxref
+            // rendered inside cssinfo) carries positions inside a synthetic intermediate
+            // string, not the markdown source — mixing its `col` with the outer span's
+            // `end_col` produces nonsense (col=23, end_col=11). Take positions from the
+            // first scope frame that has a real `line` set, and don't let later frames
+            // partially overwrite. `scope()` iterates innermost-first, so this means
+            // innermost-with-real-positions wins.
+            if issue.line == 0 && entries.line != 0 {
```
caugner, line 175: Related: Once https://github.com/mdn/rari/pull/701 lands, we might want to reuse `Sourcepos` and pass it directly on the issue instead of the four fields, if that is feasible. That would save us some destructurings.

crates/rari-doc/src/issues.rs
caugner, line 164: This could deserve some unit tests.

caugner: Tests added via 93c6aea6d15849d89f9ea53d93033bab9b49fcd3.

crates/rari-doc/src/issues.rs
```diff
@@ -56,6 +56,9 @@ pub struct IssueEntries {
     req: u64,
     ic: i64,
     col: i64,
+    /// Line number; `0` means no real markdown position (e.g. synthetic
+    /// intermediate strings from nested `render_and_decode_ref` calls). Treated
+    /// as unset in `populate_issue_from_scope` — position fields are atomic.
```
LeoMcA: Lets move a comment like this down to L177 where we do the actual comparison

crates/rari-doc/src/issues.rs
```diff
@@ -171,17 +174,14 @@ where
             if entries.req != 0 {
                 issue.req = entries.req;
             }
-            if entries.col != 0 {
-                issue.col = entries.col;
-            }
-            if entries.line != 0 {
+            // Position fields form an atomic group: mixing `col`/`end_col` from different
+            // scope frames produces nonsense. Take all four from the first frame with a
+            // real `line`; inner synthetic-string spans carry `line=0` and are skipped.
```
LeoMcA: Not sure this comment really makes sense after the change has been made, let's have a comment along the lines of "issue.line == 0 indicates synthetic values from a nested macro, reset the values to their parent's"

crates/rari-doc/src/issues.rs
```diff
@@ -647,4 +647,61 @@ mod tests {
...
+            "templ",
+            templ = "cssxref",
+            line = 0i64,
+            col = 23i64,
+            end_line = 0i64,
+            end_col = 35i64
+        );
+        let _i = inner.enter();
+
+        tracing::warn!(file = "test.md");
```
LeoMcA: Can we use a more specific name here? Claude reckons we could end up with conflicts if multiple tests emit issues all for "test.md"

LeoMcA (approved):
A few nits, otherwise looks good!

crates/rari-doc/src/issues.rs
caugner: 4f0e37006fef1744701a31903b1464b18c61ba15

crates/rari-doc/src/issues.rs
caugner: 4fbffb4eda5a9f90dbb68f9ffce309afbaf6634e

crates/rari-doc/src/issues.rs
caugner: d21eca7378eaf0046a443a2024afa01871f59b7b

PR #715, opened 2026-05-21
fix(jsxref): build Web/JavaScript index to resolve names

caugner:
### Description

Update `jsxref` to resolve names against a lazy in-memory index of all en-US `Web/JavaScript/Reference/*` page slugs, instead of two per-call `RariApi::get_page_nowarn` probes. The index:

- Indexes each page under its full sub-path (`Statements/for...of`).
- Adds aliases for the `Global_Objects/`-stripped form (`Array`, `Array/from`, `undefined`), the `Operators/*` and `Statements/*` bare leaves (`null`, `typeof`, `const`, `return`), and a namespace strip for the class-style namespaces `Intl` and `Temporal` (`Collator` → `Intl/Collator`, `Collator/compare` → `Intl/Collator/compare`).
- Looks up case-sensitively (so `Set` / `set` and `Function` / `function` are distinguishable), falling back to a precomputed lowercase map on miss to keep the case-folded fallback O(1).

When a case-sensitive miss has exactly one case-folded match, the lookup resolves but emits a new `templ-ill-cased-arg` flaw pointing at the canonical casing. When a bucket has multiple case-sensitive candidates (today: lowercase `function`, `class`, `import`, `async_function`), the resolver refuses and emits `templ-invalid-arg`.

A fragment embedded directly in the argument (e.g. `{{jsxref("Array.prototype.map#examples")}}`) is split off before normalization and the index lookup, then re-appended to the URL, so the bare name still resolves via the index. When both an embedded fragment and the separate `anchor` argument are given, the embedded fragment wins and `anchor` is ignored (emitting `templ-invalid-arg`); empty `anchor` arguments are filtered first. This mirrors `cssxref` (#682).

### Motivation

- Translated content frequently uses unqualified forms (e.g. `{{jsxref("Collator", "Intl.Collator")}}`) that the old code couldn't resolve to a canonical path and emitted as `templ-redirected-link`.
- The old per-call `get_page_nowarn` probes followed redirects silently, masking stale slugs in content and inflating `templ-redirected-link` noise (2,588 on `main`).

### Additional details

`build --grep jsxref` across en-US + all translated locales (`es,fr,ja,ko,pt-br,ru,zh-cn,zh-tw`), counting only `jsxref`-sourced issues, against current content (2026-07-16, after the prerequisite cleanups below merged):

| Source | `main` | this PR | Δ |
|---|---:|---:|---:|
| `templ-redirected-link` | 2,588 | 0 | **−2,588** |
| `templ-broken-link` | 170 | 177 | +7 |
| `templ-ill-cased-arg` *(new)* | 0 | 16 | +16 |
| `templ-invalid-arg` *(new)* | 0 | 7 | +7 |
| **jsxref total** | **2,758** | **200** | **−2,558** |

(Excludes 2 unrelated `must be a string` argument-type errors, identical on both sides.)

The remaining diagnostics are genuine and were previously masked behind redirect chains:

- `templ-ill-cased-arg` (16): wrong-case references such as `operators/import` (canonical `Operators/import`).
- `templ-invalid-arg` (7): the genuinely ambiguous `{{jsxref("function")}}` (both `Operators/function` and `Statements/function` exist); the resolver refuses rather than guessing. Each also produces a `templ-broken-link` to `/Reference/function` via the fallback. Fixed in content by mdn/content#44732 (disambiguates these 7 to `{{jsxref("Function")}}`).
- The +7 `templ-broken-link` are real stale slugs in translated content (e.g. `Gramatica_lexica`, `StopIteration`, `ParallelArray`, `Object.prototype.watch()`) that will be fixed in translated-content via mdn/translated-content#37059 (one draft PR per locale).

Relates to #672, #682 (same indexing pattern applied to `domxref` and `cssxref`).

Follow-up translated-content cleanup for the remaining stale `templ-broken-link` references surfaced above (one draft PR per locale):

- https://github.com/mdn/translated-content/issues/37059

Prerequisite translated-content cleanups (all merged):

- https://github.com/mdn/translated-content/pull/36163
- https://github.com/mdn/translated-content/pull/36164
- https://github.com/mdn/translated-content/pull/36165
- https://github.com/mdn/translated-content/pull/36166
- https://github.com/mdn/translated-content/pull/36167
- https://github.com/mdn/translated-content/pull/36168
- https://github.com/mdn/translated-content/pull/36169
- https://github.com/mdn/translated-content/pull/36170


crates/rari-doc/src/templ/js_ref_index.rs
```diff
@@ -312,6 +343,35 @@ mod tests {
         );
     }
 
+    #[test]
+    fn statement_leaf_resolves_bare() {
+        let map = fixture();
+        assert_eq!(resolve_from_map(&map, "const"), Some("Statements/const"));
+        // `for...of` still resolves via the full `Statements/` sub-path
+        // (its `.` makes a bare-leaf shortcut ambiguous with paths).
```
LeoMcA: This feels like something which should/can be fixed: we should normalize only a single `.` -> `/` but leave multiple alone, with some regex like `(?<!\.)\.(?!\.)`

crates/rari-doc/src/templ/js_ref_index.rs
```diff
@@ -21,6 +27,17 @@ const GLOBAL_OBJECTS_PREFIX: &str = "Global_Objects/";
...
+/// Namespaces whose direct children are *classes* (`Intl.Collator`,
+/// `Temporal.Instant`, etc.) and should be addressable without the
+/// namespace prefix.
+///
+/// Excludes static-API namespaces like `Reflect`, `Atomics`, `Math`, and
+/// `JSON` whose children are *methods* (e.g. `Reflect.set`, `Math.sin`).
+/// Stripping the namespace there would alias `set`/`sin`/etc. to a JS
+/// reference page, creating spurious collisions with the global `Set`
+/// class, the `Functions/set` getter syntax, etc.
+const NAMESPACE_PREFIXES: &[&str] = &["Global_Objects/Intl/", "Global_Objects/Temporal/"];
```
LeoMcA, line 39: not an enormous fan of hard-coding, but this could be a follow-up fix/enhancement: could we come up with a heuristic for these: perhaps checking the pagetype of both parent and child?

crates/rari-doc/src/templ/js_ref_index.rs
```diff
@@ -0,0 +1,492 @@
...
+        let idx = fixture();
+        // Caller normalizes `Array.from` → `Array/from`.
+        assert_eq!(
+            resolve_from_index(&idx, "Array/from"),
+            Some("Global_Objects/Array/from")
+        );
+        // Caller normalizes `Array.prototype.map` → `Array/map`.
+        assert_eq!(
+            resolve_from_index(&idx, "Array/map"),
+            Some("Global_Objects/Array/map")
```
LeoMcA: These two aren't really testing anything; perhaps this test is no longer needed with the testing of normalization in jsref, but I think I'd prefer to move normalization here, within `resolve_js_ref` - it feels like something the index should own, as changes to the indexing may affect normalization and vice versa. What do you think?

LeoMcA (approved):
Overall looks good! A handful of things I noticed:

crates/rari-doc/src/templ/js_ref_index.rs
caugner: Fixed in b4ce5c1c: normalization now converts only a lone `.`, leaving runs intact, so `for...of` resolves by its bare leaf. Done as a char scan rather than a lookaround regex, since Rust's `regex` crate doesn't support lookaround.

crates/rari-doc/src/templ/js_ref_index.rs
caugner: Good call. Moved normalization into `resolve_js_ref` in b4ce5c1c so the index owns it, and reworked these tests to feed raw names (`Array.prototype.map`) through `normalize` so they actually exercise it.

crates/rari-doc/src/templ/js_ref_index.rs
caugner: Agreed this is a good follow-up rather than a blocker. The hard-coded list is pragmatic: only `Intl` and `Temporal` have class-style children today, so the payoff is mostly future-proofing. A page-type heuristic is feasible (strip the namespace when the parent is `JavascriptNamespace` and the child is `JavascriptClass`, which naturally excludes `Math`/`JSON`/`Reflect` with their method/property children), so I've filed it as a follow-up: #800.

PS: `Intl` and `Temporal` are also less collision-prone because their members have quite distinct names.

LeoMcA (approved):

LeoMcA (approved):

PR #717, opened 2026-05-22
feat(cli): add `--grep` build option

caugner:
### Description

Update `rari build` to accept a `--grep <NEEDLE>` flag:

- Walks `content_root` (and `content_translated_root` when configured) using the existing `markdown:index.md` type filter, and builds every `index.md` whose raw bytes contain `<NEEDLE>` (ASCII case-insensitive substring).
- Honors `--locale`: the walk visits `content_root` (en-US is always included by `finalize_requested_locales`) plus one translated subdirectory per non-en-US locale, instead of every configured root.
- Conflicts with positional `FILES` and `--file-list` (via `clap`'s `conflicts_with_all`) so the selection mechanism is unambiguous — use one or the other, never both.
- Rejects `--grep ""` at parse time via `clap::builder::NonEmptyStringValueParser`.
- Logs `nothing to build` and exits cleanly when zero files match, instead of falling through to a full rebuild.

### Motivation

Bake the existing `rg -l <NEEDLE> | rari build --file-list -` workflow into one step, primarily for rebuilding all pages that invoke a given template macro (e.g. all `{{domxref(...)}}` callers) to inspect macro-related issues.

### Additional details

- Implementation lives in [`crates/rari-doc/src/walker.rs`](https://github.com/mdn/rari/blob/build-grep/crates/rari-doc/src/walker.rs) as `pub fn grep_doc_files(needle: &str, filter: LocaleFilter<'_>) -> Result<Vec<PathBuf>, DocError>`. It reuses `walk_builder` (gitignore + `index.md` type filter), drives it with `build_parallel()`, and matches via `memchr::memmem::Finder` against an ASCII-lowercased copy of each file's bytes.
- CLI wiring in [`crates/rari-cli/main.rs`](https://github.com/mdn/rari/blob/build-grep/crates/rari-cli/main.rs) runs `--locale` validation first so the walker can scope its roots accordingly. `Cache::Dynamic` is selected automatically because the existing `arg_files`-driven branch fires.
- Files that fail to read are logged at `warn` and skipped — one unreadable file doesn't abort the whole grep.
- Unit tests in `walker.rs` cover empty-needle short-circuit, ASCII case folding, and non-ASCII byte-literal matching via a path-taking `grep_doc_files_in` helper.
- Verified against the corpus: `rg -l -i svgref … --glob 'index.md' | wc -l` = 581; `rari build --grep svgref` logs `--grep matched 581 file(s)`. Case insensitivity confirmed (`--grep SVGRef` and `--grep svgref` yield identical results).

### Out of scope

- Regex matching — `rg -l … | rari build --file-list -` remains the escape hatch.
- Grepping non-doc page types (blog / curriculum / spotlights / generics) — `arg_files` only feeds `read_docs_parallel::<Page, Doc>`, so `--grep` matches positional `FILES` scope (docs only).

_None._

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2709,6 +2709,7 @@ dependencies = [
  "inventory",
  "itertools",
  "lol_html",
+ "memchr",
  "memoize",
  "percent-encoding",
  "pretty_yaml",
--- a/crates/rari-cli/main.rs
+++ b/crates/rari-cli/main.rs
@@ -215,6 +215,7 @@ struct BuildArgs {
     #[arg(
         long,
         value_name = "NEEDLE",
+        value_parser = clap::builder::NonEmptyStringValueParser::new(),
         conflicts_with_all = ["files", "files_flag", "file_list"],
         help = "Build only docs whose Markdown source contains <NEEDLE> (ASCII case-insensitive)"
     )]
@@ -538,9 +539,6 @@ fn main() -> Result<(), Error> {
             }
 
             if let Some(needle) = args.grep.as_deref() {
-                if needle.is_empty() {
-                    return Err(anyhow!("--grep requires a non-empty <NEEDLE>"));
-                }
                 let matches = rari_doc::walker::grep_doc_files(needle)?;
                 info!("--grep matched {} file(s)", matches.len());
                 if matches.is_empty() {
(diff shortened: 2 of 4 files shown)


LeoMcA (approved):
Clean! Lgtm

PR #737, opened 2026-06-01
fix(specs): make `filename` optional

LeoMcA:
web-specs v4.0.0 drops `filename` from RFC spec nightly objects, breaking deserialization of index.json, which means `rari` won't run.

https://github.com/mdn/dex/actions/runs/26729292623
https://github.com/mdn/dex/actions/runs/26729932043

### DIFF
--- a/crates/css-syntax-types/src/specs.rs
+++ b/crates/css-syntax-types/src/specs.rs
@@ -50,7 +50,7 @@ pub struct NightlySpec {
     #[serde(rename = "alternateUrls", default)]
     pub alternate_urls: Vec<String>,
     pub repository: Option<String>,
-    pub filename: String,
+    pub filename: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
@@ -58,7 +58,7 @@ pub struct ReleaseSpec {
     pub url: String,
     pub status: String,
     pub pages: Option<Vec<String>>,
-    pub filename: String,
+    pub filename: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


LeoMcA:
Diff failure is expected, as it runs rari from `main` first. Other tests pass. It's a trivial change to fix an issue which prevents rari from starting, therefore self-approving in this exceptional circumstance.

ddbeck:
@LeoMcA I can't quite figure out why just yet, but this seems to have broken the content inventory. Maybe some sort of error is being written to stdout?

I'll have more time tomorrow or Wednesday to diagnose this better, but if you want to take a look:

https://github.com/ddbeck/mdn-content-inventory/actions/runs/26751903900/job/78894647965#step:6:32

ddbeck:
Oh wait, I got the causality backwards. This is a _fix_ for a change upstream? I'll see if the content inventory builds cleanly tomorrow and file a bug if it's still ongoing.

(Seems like rari ought to know to not download a semver major that it's not explicitly equipped to handle, then fallback to the last release that it does?)

LeoMcA:
@ddbeck yes, this fixes the issue you're seeing in CI: that run checks out https://github.com/mdn/content/commit/0b2ddf6c871a55eebe31f4d8808b8c975bfa41a8 which is prior to when the fixed rari version (via an updated fred version) got merged into content - it should succeed tomorrow!

> (Seems like rari ought to know to not download a semver major that it's not explicitly equipped to handle, then fallback to the last release that it does?)

Indeed, and we even have a mechanism for this I believe, but wanted to go with the simple approach for the hotfix - thanks for the reminder to have a look at that!

ddbeck:
Sounds good. Thank you for the quick fix and response, @LeoMcA!

PR #752, opened 2026-06-05
fix(settings): default BUILD_OUT_ROOT to ./build

### LINKED ISSUE #733: Default `BUILD_OUT_ROOT` to a gitignored path inside the workspace
### Current behavior

`BUILD_OUT_ROOT` has no default. If it isn't set in env, `.config.toml`, or `~/.config/rari/config.toml`, `cargo run -- build` panics with `.expect("No BUILD_OUT_ROOT")` ([`crates/rari-doc/src/build.rs:86`](https://github.com/mdn/rari/blob/main/crates/rari-doc/src/build.rs#L86)). The getter itself returns `Err(EnvError::NoBuildOut)` ([`crates/rari-types/src/globals.rs:43-49`](https://github.com/mdn/rari/blob/main/crates/rari-types/src/globals.rs#L43-L49)) and the field is `Option<PathBuf>` ([`crates/rari-types/src/settings.rs:66`](https://github.com/mdn/rari/blob/main/crates/rari-types/src/settings.rs#L66)).

### Proposed change

In `Settings::new_internal` ([`crates/rari-types/src/settings.rs`](https://github.com/mdn/rari/blob/main/crates/rari-types/src/settings.rs#L133)), default `build_out_root` to `./build` (CWD-relative) when none of the config sources provide it. Add `/build` to [`.gitignore`](https://github.com/mdn/rari/blob/main/.gitignore).

### Motivation

- Removes a mandatory config step for new contributors — `cargo run -- build` works out of the box (with `CONTENT_ROOT` set).
- CWD-relative resolution gives **per-worktree separation by default**: running rari from a separate `git worktree` of the repo writes to a worktree-local `build/` rather than contending on a shared path like `/tmp/rari`. Useful when running parallel builds across worktrees.
- The result stays inside the rari repo for the typical `cargo run -- build` invocation, and `/build` in `.gitignore` keeps it out of commits.

### Trade-off

For users invoking an installed `rari` binary from a content checkout, the binary will now silently create `./build` in their CWD instead of panicking. Previously they had to opt in via env or config. Worth a release note.

reuteelf:
default BUILD_OUT_ROOT to ./build

### Description

Use CWD-relative build directory when BUILD_OUT_ROOT is not provided in the config files. 
Also add /build to .gitignore.

Fixes #733


caugner (approved):
LGTM, this is a reasonable improvement. Tested locally with `cargo run build --grep glossary --locale en-us`.

Issue #213, opened 2025-05-12
I can't run content server on my local machine...


Zorono:
### What information was incorrect, unhelpful, or incomplete?

this is happening in **localhost:8080**
```
Oops! There seems to be an issue loading this page:


http://localhost/en-US/docs/Learn_web_development/Core/CSS_layout/Responsive_Design

Because

Loop request blocked


Your details are below:
User: -   Group: group1   IP: 127.0.0.1


[Go Back to previous page](javascript:history.back())
```

I am trying to run content server on my local machine to browse the documentations locally... but it fails to load any page with the mentioned error page...

Steps:
`yarn install` firstly and then

1. `yarn run info:rari && env CONTENT_ROOT=files BUILD_OUT_ROOT=build npx rari-build`
2.  `yarn run info:rari && yarn up-to-date-check && env CONTENT_ROOT=files REACT_APP_DISABLE_AUTH=true BUILD_OUT_ROOT=build npx rari-server` : 
```
🐥 This command is now using rari: https://github.com/mdn/rari
🐞 Please report any issues here:  https://github.com/mdn/rari/issues/new?template=bug.yml

(node:2217986) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
[server] (node:2218216) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
[server] (Use `node --trace-warnings ...` to show where the warning was created)
[server] [HPM] Proxy created: /  -> http://localhost:8080
[server] [HPM] Proxy created: /  -> https://developer.mozilla.org
[server] [HPM] Proxy created: /  -> https://developer.allizom.org
[server] [HPM] Proxy created: /  -> https://mdn.github.io/shared-assets/
[server] [HPM] Proxy rewrite rule created: "^/shared-assets/" ~> "/"
[server] 
[server] 🗑️  This command is deprecated, and will be removed soon.
[server] 
[server] CONTENT_ROOT: /sdb1/Storage/dump/mdn@content/files 
[server] Listening on port 5042
[server] Your EDITOR is set to: /usr/bin/nano
```

but **localhost:5042** results in:
```
[server] using http://localhost:8083/en-US/
[rari]  WARN rari_doc::cached_readers: No generic content config found ignore=true
[rari]  WARN rari_doc::pages::types::spa_homepage: page not found /en-US/blog/javascript-temporal-is-coming/ (BlogPost)
[...]

PR #760, opened 2026-06-15
fix(utils): centralize git calls and panic on non-zero exit codes

pzhlkj6612:
### Description

Hi. This PR will make failed git commands be discovered by checking git's exit code.

### Motivation

The `std::io::Result::expect()` call only cares about failed launch of programs, not the result of those programs. The missing check has already made debugging hard. Checking the exit code and preserving stderr will help developers to see if git related things go wrong.

My test with [mdn/content at 0985f662bf6c2351278ddb6c6629827d00ff6755](https://github.com/mdn/content/tree/0985f662bf6c2351278ddb6c6629827d00ff6755)):

Before (100b9cf245b6fdce5744055de7cd5cd6c9212de4):

```console
# npm run start
...
[rari] Rari server started on http://0.0.0.0:8083
...
[rari] thread 'main' panicked at crates/rari-doc/src/pages/types/spa_homepage.rs:118:10:
[rari] failed to execute process: Os { code: 2, kind: NotFound, message: "No such file or directory" }
[rari] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After (this):

```console
[rari] thread 'main' panicked at crates/rari-doc/src/pages/types/spa_homepage.rs:133:6:
[rari] failed to execute git rev-parse: CommandFailed { exit_code: "128", stderr: "fatal: detected dubious ownership in repository at '/src'\nTo add an exception for this directory, call:\n\n\tgit config --global --add safe.directory /src\n" }
[rari] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Additional details

Replacing the existing `exec_git()` function in "crates/rari-tools/src/git.rs" [^git-rs] with the implementation of the newly added inner function `exec_git()` is preferred, but it will be a large cross-module change, so I didn't make it. Discussion is needed.

[^git-rs]: https://github.com/mdn/rari/blob/100b9cf245b6fdce5744055de7cd5cd6c9212de4/crates/rari-tools/src/git.rs#L23-L29

I'm not a Rust developer and I just discovered that this can be improved. Please review my code and point me out if I can make it more Rust-ify. Thank you in advance.

- Relates to https://github.com/mdn/rari/issues/213 (moved from https://github.com/orgs/mdn/discussions/795).
- Relates to https://github.com/mdn/content/issues/40595.


crates/rari-doc/src/pages/types/spa_homepage.rs
```diff
@@ -96,26 +106,46 @@ fn recent_contributions_from_git(
     path: &Path,
     repo: &str,
 ) -> Result<Vec<HomePageRecentContribution>, DocError> {
-    let output = Command::new("git")
-        .args(["rev-parse", "--show-toplevel"])
-        .current_dir(path)
-        .output()
-        .expect("failed to execute git rev-parse");
+    fn exec_git(args: &[impl AsRef<OsStr>], root: impl AsRef<Path>) -> Result<Output, GitError> {
```
LeoMcA: Let's fold this behaviour into `crates/rari-tools/src/git.rs`, and then move that to `crates/rari-utils/src/git.rs` (otherwise we'll end up with a circular dependency).

To keep the function signature the same, it's probably best to continue to panic in the `exec_git*` functions, but move their implementation to a `try_exec_git` function similar to this one, which does return a `Result`.

LeoMcA (changes requested):
Sorry about the delay here, your rust code looks great!

> Replacing the existing exec_git() function in "crates/rari-tools/src/git.rs" [1](https://github.com/mdn/rari/pull/760#user-content-fn-git-rs-6ef27c3b6df2b7db9266e6cf0514c4f3) with the implementation of the newly added inner function exec_git() is preferred, but it will be a large cross-module change, so I didn't make it. Discussion is needed.

Thanks for checking first, but yes - absolutely, let's unify the code paths and ensure we fix the bug for all of them.

crates/rari-doc/src/pages/types/spa_homepage.rs
pzhlkj6612: Thanks for your review! Please see 2f1a3b01b781d9cd880bd2cb8ab9e3ad88836032 .

To make PR's history readable, I didn't do force-pushing, though.

crates/rari-utils/src/git.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/// Run `git` with `args` with `root` as current directory.
+/// For tests this will run the first arg as command, eg.:
+/// Instead of `git mv foo bar` -> `mv foo bar`.
+pub fn exec_git_with_test_fallback(args: &[impl AsRef<OsStr>], root: impl AsRef<Path>) -> Output {
+    let git = OsStr::new("git");
+    let echo = OsStr::new("echo");
+
+    let (command, args) = if cfg!(test) {
```
LeoMcA: Ah, looks like this function needs to remain in `rari-tools` actually - the `cfg!(test)` doesn't work across crate boundaries, so all the tests in `rari-tools` end up failing because they try real git commands. I think it should be fine to just move this function back, and keep everything else here

LeoMcA (changes requested):

crates/rari-utils/src/git.rs
pzhlkj6612: I admit that I didn't read CI workflow and ran any `cargo test ...`, but built a release binary and tested with MDN content.

Now everything should work fine. See 96d43c202e29e63bf42fd1c5b1a02a26c9bc8148.

crates/rari-utils/src/git.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Debug, Error)]
+pub enum GitError {
+    #[error("failed to run git: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("git command exited with non-zero code {exit_code}: {stderr}")]
+    CommandFailed { exit_code: String, stderr: String },
+}
+
+pub fn exec_git(args: &[impl AsRef<OsStr>], root: impl AsRef<Path>) -> Output {
+    try_exec_git("git", args, root).expect("failed to execute process")
```
LeoMcA: I think "failed to execute process" implies git failed to even start, so perhaps something like this:

```suggestion
    try_exec_git("git", args, root).expect("git command failed")
```

crates/rari-tools/src/git.rs
```diff
@@ -17,25 +19,5 @@ pub fn exec_git_with_test_fallback(args: &[impl AsRef<OsStr>], root: impl AsRef<
...
-    command: impl AsRef<OsStr>,
-    args: &[impl AsRef<OsStr>],
-    root: impl AsRef<Path>,
-) -> Output {
-    Command::new(command)
-        .args(args)
-        .current_dir(root)
-        .output()
-        .expect("failed to execute process")
+    try_exec_git(command, args, root).expect("failed to execute process")
```
LeoMcA: Each of the call sites of this function were previously checking `!output.status.success()` themselves:
https://github.com/mdn/rari/blob/96d43c202e29e63bf42fd1c5b1a02a26c9bc8148/crates/rari-tools/src/move.rs#L166-L171
https://github.com/mdn/rari/blob/96d43c202e29e63bf42fd1c5b1a02a26c9bc8148/crates/rari-tools/src/remove.rs#L182-L187
https://github.com/mdn/rari/blob/96d43c202e29e63bf42fd1c5b1a02a26c9bc8148/crates/rari-tools/src/remove.rs#L195-L200
https://github.com/mdn/rari/blob/96d43c202e29e63bf42fd1c5b1a02a26c9bc8148/crates/rari-tools/src/sync_translated_content.rs#L286-L292

There's a few approaches which we could take here, but I think best is to ensure this function returns a `Result<Output, GitError>`, and return simply:

```suggestion
    try_exec_git(command, args, root)
```

And then in each of those above places, `.map_err` rather than check `!output.status.success()` - keeping the old error message, e.g. for the first example, something like:

```rust
    exec_git_with_test_fallback(
        &[
            OsStr::new("mv"),
            old_folder_path.as_os_str(),
            new_folder_path.as_os_str(),
        ],
        root_for_locale(locale)?,
    ).map_err(|e| ToolError::GitError(format!("Failed to move files: {e}")))?;
```

LeoMcA (changes requested):
> I admit that I didn't read CI workflow and ran any cargo test ..., but built a release binary and tested with MDN content.

ha, no problem - the CI usually catches these errors for me, but I need to manually run it for you because it's your first contribution.

Good changes, thanks - we're very nearly there! Just a small bit of clean-up/cosmetics, really:

crates/rari-tools/src/git.rs
pzhlkj6612: > Each of the call sites of this function were previously checking `!output.status.success()` themselves:
> 
> ...
> 
> There's a few approaches which we could take here, but I think best is to ensure this function returns a `Result<Output, GitError>`, and return simply:
> 
> And then in each of those above places, `.map_err` rather than check `!output.status.success()` - keeping the old error message, e.g. for the first example, something like:
> 
> ...

Done, cf08f6c625611a33ed35ad97c974fd9e80dbfb44 . I didn't examine the code surrounding those changed lines, then left those checks meaningless.

crates/rari-utils/src/git.rs
pzhlkj6612: > I think "failed to execute process" implies git failed to even start, so perhaps something like this:

Well, I don't quite understand the relationship between the reason and the suggestion you wrote, but I think using "git command failed" is just more informative, so, 13253aaffa4cb790642130fb9232011ec0c0f0e9 .

Could you please elaborate?

LeoMcA: "failed to execute process" made sense when we only panicked when the git command was missing etc. but now we're also panicking when git returns an error it makes sense to use a message which covers both - "git command failed" could reasonably both mean that the git command failed to start, and that it started and then failed

LeoMcA (approved):
Great stuff, thank you!

crates/rari-utils/src/git.rs
pzhlkj6612: Ah, now I understand. Thank you!

PR #774, opened 2026-07-01
fix(curriculum): preserve frontmatter when fixing flaws

### LINKED ISSUE #773: `fix-flaws --curriculum` strips frontmatter
See:

- https://github.com/mdn/curriculum/pull/126
- https://github.com/mdn/curriculum/pull/126/changes/9837cb1bc19915487fdc498baf435e7e32b24e7c

caugner:
### Description

Fix `content fix-flaws --curriculum` to preserve the frontmatter and H1 title when rewriting links in curriculum pages.

- Make `Curriculum` store the full file contents plus a body offset, mirroring `Doc`, so `raw_content()` returns the full file (frontmatter + H1 + body) while `content()` still returns the stripped body.
- Set `fm_offset()` to the prefix line count so issue line numbers and `raw_content()` share the same coordinate space.

### Motivation

Prevent `fix-flaws --curriculum` from corrupting curriculum files. `Curriculum` was the only page type whose `raw_content()` returned the body with frontmatter and the H1 already stripped; since `fix_page` writes `raw_content()` back to disk, the command truncated every touched curriculum file to its body, discarding the frontmatter and H1.

### Additional details

The fix keeps the fix pipeline (`fix_page`, `collect_suggestions`, `apply_suggestions`) unchanged — only `Curriculum`'s representation changes. Verified end-to-end against the local curriculum checkout: `fix-flaws --curriculum` rewrote 19 files' redirected links while leaving frontmatter and H1 fully intact.

Fixes #773.


crates/rari-doc/src/pages/types/curriculum.rs
```diff
@@ -534,4 +550,42 @@ mod tests {
...
+        let raw = "---\ntitle: Test\n---\n\n# My Title\n\nBody with a [link](/en-US/docs/foo).\n";
+        let (fm, title, content_start) = parse_curriculum(raw).unwrap();
+        assert_eq!(fm, "title: Test");
+        assert_eq!(title, "My Title");
+        // `content()` returns the stripped body (frontmatter and H1 removed).
+        assert_eq!(
+            &raw[content_start..],
+            "\nBody with a [link](/en-US/docs/foo).\n"
+        );
+        // `raw_content()` returns the full file: the stripped prefix + body reconstructs it.
```
LeoMcA: This comment doesn't match what the next line checks. Since we're not actually testing the methods, but testing that slicing at `content_start` produces the correct content, perhaps we just bin the comments (this one, the one above, and the one below)?

crates/rari-doc/src/pages/types/curriculum.rs
```diff
@@ -534,4 +550,42 @@ mod tests {
...
+        assert_eq!(
+            &raw[..content_start],
+            "---\ntitle: Test\n---\n\n# My Title\n"
+        );
+        // `fm_offset()` counts the prefix lines (frontmatter + blank line + H1).
+        assert_eq!(raw[..content_start].lines().count(), 5);
+    }
+
+    #[test]
+    fn test_parse_curriculum_body_matches_legacy_stripping() {
```
LeoMcA: I don't think we really need to/should merge this test with the legacy code in it - it won't make sense in the future - can we parameterize the above test with the examples used in this one?

LeoMcA (approved):
Looks good: a couple of thoughts about the tests:

crates/rari-doc/src/pages/types/curriculum.rs
caugner: I've merged them into a parameterized test via [e6f7e38](https://github.com/mdn/rari/pull/774/commits/e6f7e38b903f7d0487b1e0d1b11f177c730e0d6f), but kept one comment per test case to make it easier to understand what each of them tests.

crates/rari-doc/src/pages/types/curriculum.rs
caugner: Removed.

PR #777, opened 2026-07-01
fix(issues): report 1-based line/column consistently

### LINKED ISSUE #776: Inconsistent line/column base between comrak- and tree-sitter-sourced issues
## Summary

The internal `Issue` struct mixes two position conventions depending on which parser produced the issue, with no normalization:

| Source | Origin | line base | col base |
|---|---|---|---|
| `broken-link` / `redirected-link` / `ill-cased-link` | comrak `data-sourcepos` (`fix_link.rs`) | **1-based** | **1-based** |
| `templ-*` | tree-sitter `start_position()` (`templ/render.rs`) | **0-based** | **0-based** |

For a link on file line `N` (`fm_offset` = frontmatter line count):

- broken-link: `comrak_line + fm_offset` = `(N - fm_offset) + fm_offset` = **N** (1-based)
- templ: `tree_sitter_row + fm_offset` = `(N - 1 - fm_offset) + fm_offset` = **N - 1** (0-based)

So the two sources are consistently off from each other by exactly 1, in both line and column.

## Impact

Downstream consumers can't apply a single convention. A tool that adds `+1` to convert to 1-based positions (correct for the tree-sitter `templ-*` issues) shifts every comrak-sourced `broken-link` by one on both line and column.

Rari's own output shows the mirror symptom: `DisplayIssue::from_issue` (`issues.rs`) uses `issue.line` verbatim (templ lines end up 1 too low) and does `byte_to_char_column(...) + 1` on the column assuming 0-based (broken-link columns end up 1 too high).

## Example

For a `broken-link` on line 135, `--issues` emits `line: 135, col: 63, end_col: 122` — all correctly 1-based. A `templ-broken-link` on the same line would emit `line: 134` and 0-based columns.

## Suggested fix

Normalize both sources to a single convention at emission (e.g. subtract 1 from the comrak line/col in `fix_link.rs` so everything is 0-based like tree-sitter), then audit the `DisplayIssue` and LSP conversions, which currently compensate inconsistently.

caugner:
### Description

Normalize issue positions to 1-based line and column across all sources:

- `templ-*` issues (tree-sitter, previously 0-based) now add 1 to line and column in `render`.
- `broken-link`/`redirected-link`/`ill-cased-link` (comrak sourcepos) are already 1-based and unchanged.
- `DisplayIssue::from_issue` subtracts 1 before the byte-to-character conversion to match the now-1-based byte columns.

### Motivation

Tree-sitter and comrak used different position bases, so downstream consumers of `issues.json` couldn't apply a single convention: a uniform `+1` to reach 1-based (correct for `templ-*`) shifted every `broken-link` by one on both line and column. This also fixes a latent over-count of `broken-link` columns in `DisplayIssue`/LSP output.

### Additional details

Verified against real content — `templ-broken-link` in `glossary/mixin` now reports `line 8, col 247` (was `line 7, col 246`), matching the true source position; `broken-link` in `firefox/releases/62` stays `line 135, col 63`. `DisplayIssue` columns and `sourceContext` highlighting are now correct for both sources.

Fixes #776.


crates/rari-doc/src/templ/render.rs
```diff
@@ -71,10 +71,15 @@ pub(crate) fn render(env: &RariEnv, input: &str, offset: usize) -> Result<Render
...
                 let ident = &mac.ident;
                 let name = ident.to_ascii_lowercase();
-                let line = i64::try_from(mac.pos.0 + offset).unwrap_or(-1);
-                // mac.pos.1 is in bytes from tree-sitter
-                let col = i64::try_from(mac.pos.1).unwrap_or(-1);
-                // Calculate end_col in bytes: start byte + length of macro in bytes
+                // tree-sitter positions are 0-based; add 1 to report 1-based
+                // positions, consistent with comrak's sourcepos (see `fix_link`).
+                let line = i64::try_from(mac.pos.0 + offset)
+                    .map(|l| l + 1)
```
LeoMcA: Verified running against the tests below, why don't we do:


```suggestion
                let line = i64::try_from(mac.pos.0 + offset + 1)
```

crates/rari-doc/src/templ/render.rs
```diff
@@ -71,10 +71,15 @@ pub(crate) fn render(env: &RariEnv, input: &str, offset: usize) -> Result<Render
...
+                    .unwrap_or(-1);
+                // mac.pos.1 is a 0-based byte column from tree-sitter.
+                let col = i64::try_from(mac.pos.1).map(|c| c + 1).unwrap_or(-1);
```
LeoMcA: Similarly:

```suggestion
                let col = i64::try_from(mac.pos.1 + 1).unwrap_or(-1);
```

LeoMcA (approved):
Looks good, but should be possible to collapse a couple of lines:

PR #808, opened 2026-07-24
fix(css_info): don't translate the initial value `all`

naimonon77:
## Description

Fixes incorrect localization of the CSS keyword `all` in the initial value for `transition-property` on localized pages.

### Problem
On localized versions (e.g., Japanese), the "Initial value" for `transition-property` was being displayed as a localized string (e.g., "すべて") instead of the CSS keyword `all`. This occurs because the `CSSInfo` helper was localizing all initial values that had entries in the localization data, without checking if the value is a reserved CSS keyword that should never be translated.

### Solution
Modified the condition in `write_computed_output` to explicitly exclude `all` from localization. The keyword is now always rendered as `<code>all</code>` regardless of locale, while preserving the existing behavior for other initial values.

### Changes
- **File**: `crates/rari-doc/src/helpers/css_info.rs`
- Added condition `s == "all"` to prevent localization of the keyword
- Added unit test `renders_keyword_initial_value_as_code_for_locales` to verify behavior

## Testing

> [!NOTE]
> Verify the fix prevents localization of the CSS keyword `all` in the initial value for `transition-property` on localized pages.

### Build Japanese locale content

```bash
cargo run -- build --content --no-basic --locale ja --issues-per-locale {translated-content-path}/files/ja/web/css/reference/properties/transition/index.md
```

### Verify the output

```bash
jq -r '.doc.body[] | select(.value.title=="公式定義") | .value.content' /tmp/rari/ja/docs/web/css/reference/properties/transition/index.json | grep -e ": <code>all</code>" -e ": すべて"
```

> [!IMPORTANT]
> Expected output: `: <code>all</code>`
> If you see `: すべて` instead, the fix is not working correctly.

### Screenshot

Before:
image

<br>

After:
image

## Related Issues
- https://github.com/mdn/rari/issues/318
- Reported in [mdn/translated-content#28897](https://github.com/mdn/translated-content/issues/28897) (Japanese)


crates/rari-doc/src/helpers/css_info.rs
```diff
@@ -186,7 +186,9 @@ pub fn write_computed_output(
                     .get(&s[1..s.len() - 1])
                     .unwrap_or(&Value::Null);
                 return write_computed_output(env, out, locale, s_data, property, at_rule);
-            } else if property == "initial" && !mdn_data_files().css_l10n.contains_key(s) {
+            } else if property == "initial"
+                && (s == "all" || !mdn_data_files().css_l10n.contains_key(s))
```
LeoMcA: Let's do the inverse, instead of excluding `"all"`, let's check if the value is a known value for `initial` which should be l10ned, something like:

```rust
const INITIAL_L10N_VALUES = [
    "\"\"",
    "\". \"",
    "autoForSmartphoneBrowsersSupportingInflation",
    "dependsOnUserAgent",
    "noPracticalInitialValue",
    "noneButOverriddenInUserAgentCSS",
    "seeProse",
    "startOrNamelessValueIfLTRRightIfRTL",
    "zoomForTheTopLevelNoneForTheRest",
];
```

and then:

```rust
} else if property == "initial" && !INITIAL_L10N_VALUES.contains(&s.as_str()) {
    return Ok(write!(out, "<code>{s}</code>")?);
```

crates/rari-doc/src/helpers/css_info.rs
```diff
@@ -347,3 +349,27 @@ fn remove_me_replace_placeholder(s: &str, replacements: &[&str]) -> String {
...
+#[cfg(test)]
+mod tests {
+    use rari_types::RariEnv;
+    use rari_types::locale::Locale;
+    use serde_json::json;
+
+    use super::write_computed_output;
+
+    #[test]
+    fn renders_keyword_initial_value_as_code_for_locales() {
```
LeoMcA: perhaps we extract the setup here to a helper and also test one of the `INITIAL_L10N_VALUES` goes through the translate branch?

LeoMcA (changes requested):
Thanks for digging into this: the shape of `mdn-data` is a bit of a mess, so this isn't straightforward.

I think the best approach would be to effectively invert the "is it all" check, with a "is it one of the known strings we localize for intitial values:

naimonon77:
Thank you for the review!
I have fixed the points you pointed out.

crates/rari-doc/src/helpers/css_info.rs
naimonon77: Thank you for the review!
I have fixed the points you pointed out.

crates/rari-doc/src/helpers/css_info.rs
naimonon77: I have fixed the points you pointed out.
I apologize if the edits do not match your intentions.

LeoMcA (approved):
Looks great, thank you!

PR #816, opened 2026-07-31
feat(baseline): add discouraged icons to sidebars, expose single status key

LeoMcA:
Depends on a content PR adding the l10n keys: https://github.com/mdn/content/pull/45017

Adds discouraged icons to the sidebars. Doing so requires calculating the discouraged statuses in rari - something the previous PR in the stack defers to fred. So, instead of calculating this in both rari and fred, this PR exposes the result of the calculation on `status.baseline` in the `index.json` which fred can consume. This field is outside of the `baseline` object to provide forwards compatibility with the mocked banner work (will be the next PR in this stack).

### DIFF
--- a/crates/rari-data/src/baseline.rs
+++ b/crates/rari-data/src/baseline.rs
@@ -24,6 +24,23 @@ pub struct Baseline<'a> {
     pub feature: &'a FeatureData,
 }
 
+#[derive(Serialize, JsonSchema, Debug, Clone, Copy, PartialEq, Eq, strum::Display)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum BaselineStatus {
+    High,
+    Low,
+    Limited,
+    Discouraged,
+    Removing,
+}
+
+impl BaselineStatus {
+    pub fn is_discouraged(self) -> bool {
+        matches!(self, Self::Discouraged | Self::Removing)
+    }
+}
+
 #[derive(Deserialize, Serialize, Debug, Clone, JsonSchema)]
 pub struct DeveloperSignals {
     pub url: String,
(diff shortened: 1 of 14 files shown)


crates/rari-data/src/baseline.rs
```diff
@@ -175,9 +210,10 @@ impl WebFeatures {
                 })
                 .collect::<Vec<_>>();
 
-            let asterisk = if sub_status
-                .iter()
-                .all(|baseline| baseline == &Some(status_for_key.baseline))
+            let asterisk = if baseline_status(feature, status_for_key).is_discouraged()
+                || sub_status
+                    .iter()
+                    .all(|baseline| baseline == &Some(status_for_key.baseline))
```
globau: `asterisk` now means two different things: "sub-keys disagree on baseline level" and "isn't discouraged/removing."

Folding the discouraged check into this field changes its semantics for a presentation concern rather than a data concern - nothing in rari itself reads `.asterisk`, it's only serialized into `index.json`.

Since fred already receives `status.baseline` separately, push this suppression into fred's rendering logic (`if status.baseline.is_discouraged()`) and leave `asterisk` here reporting only support divergence.

globau:
Follow-up work:

Once fred fully consumes `status.baseline` we should audit whether it still reads `feature.discouraged` directly to classify discouraged-vs-removing (via `discouraged.is_some()` / `removal_date.is_some()`) rather than for its descriptive copy (`reason_html`, `alternatives`, `according_to`).

If nothing depends on the raw classification we stop serialising `removal_date` at least - right now it's the one field whose presence is itself a duplicate encoding of `status.baseline`.

globau (changes requested):

globau (approved):
lgtm

PR #818, opened 2026-07-31
feat(baseline): hide deprecated banner/icon when we have a baseline one

LeoMcA:
Needs to be released after we start rendering icons/banners in fred, otherwise we'll lose any kind of deprecated indication.


globau (approved):
lgtm

PR #819, opened 2026-08-04
feat(baseline): mock discouraged banner and sidebar icon

LeoMcA:
For pages in Web/ (but not Web/Accessibility) and WebAssembly/ sections - the sections which currently have Baseline banners - replace deprecated banners with a mocked discouraged baseline status, so fred can render a mocked discouraged baseline banner.

This is (hopefully) a temporary effort, as the web-features project has more and more of these discouraged BCD keys added to their data. Because of that:
- We're not as DRY as we could be, but exposing yet another `get_baseline_status...` variant for a different set of arguments seemed excessive
- Mocked banners lack the `Support` data necessary to render browser check/crosses - I prototyped this, but it required an enormous amount of BCD parsing we currently aren't doing in rari, and this data is far less important for these banners (and always exists on the BCD table at the bottom of page if users need it)

### DIFF
--- a/crates/rari-doc/src/baseline.rs
+++ b/crates/rari-doc/src/baseline.rs
@@ -6,6 +6,7 @@
 use std::sync::LazyLock;
 
 use rari_data::baseline::{Baseline, BaselineStatus, WebFeatures};
+use rari_types::fm_types::FeatureStatus;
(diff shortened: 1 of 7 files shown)


globau (approved):
lgtm

Issue #503, opened 2026-01-29
Replace inline status macros (Deprecated_Inline, Experimental_Inline, Non-standard_Inline)


caugner:
The Web APIs docs uses inline status macros to indicate to users whether a listed feature is deprecated, experimental, or non-standard (see [here](https://github.com/search?q=repo%3Amdn%2Fcontent+%2F%28Deprecated%7CExperimental%7CNon-standard%29_Inline%2F&type=code)).

Currently, this requires a scheduled workflow run by @OnkarRuikar that syncs the content with BCD releases, by adding or removing inline macros where necessary. (Note: It also manages the `status:` frontmatter value.)

In [this comment](https://github.com/mdn/content/pull/42900#issuecomment-3813359260), Onkar cites the [sync_bcd.yml workflow in OnkarRuikar/temp](https://github.com/OnkarRuikar/temp/blob/main/.github/workflows/sync_bcd.yml) that runs daily, executes the minified script [in this Gist](https://gist.github.com/OnkarRuikar/0b332d3dcc31ea380e128005b5c1f65d) applying the changes, and opens a content PR.

While this workflow works, it updates the content, and therefore requires manual changes in translated-content to reflect the updates statuses.

Overall, it seems better if this was more dynamically handled in Rari itself.

One solution I have in mind is to add a `StatusInline` macro that takes a BCD key as parameter:

```
{{StatusInline('api.ShadowRoot.referenceTarget')}}
```

The advantage of this is that this creates a clear link between an item in content and a BCD feature, and would allow to show Baseline statuses in the feature, without requiring content updates.

PS: Although the `ReadOnlyInline` parameter does not indicate a status, and is mostly stable, one could imagine introducing an `APIStatusInline` macro that adds both the read-only indicator (if deducible from IDL), and the status indicators. However, this wouldn't work for non-standard features, because they don't appear in any IDL.

Rumyra:
I'm happy for this to be the case for non-standard but experimental and deprecated will be replaced with baseline status

See https://github.com/mdn/mdn/issues/834

LeoMcA:
Showed Ruth a prototype of Baseline icons following every link, and discussed it. The decision is:

- Don't show Widely Available icons, but show all others
- Only show icons in lists: sidebar, ol, ul, dl, not in prose
- These icons should hide other icons (e.g. if there's a dl with a link then an inline macro the link icon hides the inline macro)

PR #832, opened 2026-08-13
feat(baseline): mock baseline icon in deprecated_inline

LeoMcA:
Relates to https://github.com/mdn/fred/issues/654 and https://github.com/mdn/rari/issues/503

Rendering next to links is a bit of a larger change - one I'm working on - but as an interim fix, to remove the inline bin icons, I've extended the mock logic here too: if a deprecated_inline macro is used in a section which gets mocked banners, it gets a mocked baseline discouraged icon instead of a bin.

Before:

image

After:

image

### DIFF
--- a/crates/rari-doc/src/baseline.rs
+++ b/crates/rari-doc/src/baseline.rs
@@ -43,7 +43,7 @@ pub(crate) fn get_baseline<'a>(browser_compat: &[String]) -> Option<Baseline<'a>
 
 pub(crate) fn is_mocked_banner_section(slug: &str) -> bool {
     // only mock banners under Web and WebAssembly, as those are the sections we have Baseline banners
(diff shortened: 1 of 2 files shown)


crates/rari-doc/src/baseline.rs
```diff
@@ -41,20 +41,23 @@ pub(crate) fn get_baseline<'a>(browser_compat: &[String]) -> Option<Baseline<'a>
     None
 }
 
+pub(crate) fn is_mocked_banner_section(slug: &str) -> bool {
+    // only mock banners under Web and WebAssembly, as those are the sections we have Baseline banners
+    // don't mock under Web/Accessibility, as we don't
```
globau: incomplete comment

globau (approved):

PR #838, opened 2026-08-18
fix(mdn-data): surface missing mdn-data CSS property/at-rule entries as flaws

### LINKED ISSUE #821: Surface mdn/data issues as flaws
https://github.com/mdn/data/commit/c840ba36 added `link-parameters` and was referenced by a content page (https://github.com/mdn/content/commit/35cd8b78) before it was available in an `mdn/data` release.

This wasn't surfaced to author, and we ending up with `Value not found in DB!` in the published page.

BHUVANSH855:
Fixes #821

## Root cause

`{{CSSInfo}}` (`cssinfo()` in `crates/rari-doc/src/templ/templs/cssinfo.rs`)
looks up the property/at-rule descriptor in the locally cached `mdn-data`
snapshot. That snapshot is pulled from a published npm release
(`crates/rari-deps/src/npm.rs`), not from `mdn/data`'s git HEAD, so there's
always a window between a data commit landing and it reaching a rari build.

When the lookup misses, `cssinfo()` wrote a red "Value not found in DB!"
placeholder and returned `Ok(...)`. Because it never returned `Err`, the
problem never reached rari's existing issue-tracking (`crate::issues`),
so it shipped to production with zero record anywhere.

## Fix

`cssinfo()` now emits `tracing::warn!(source = "mdn-data-missing", ...)`
before falling back to the same placeholder, so the build issue is
recorded (bucketed under `flaws.mdn_data`, similar to how `broken_links`
and `macros` flaws are bucketed) with file/line/col pointing at the
`{{CSSInfo}}` macro call. `crates/rari-doc/src/issues.rs` gets a matching
`IssueType::MdnDataMissing` / `DIssue::MdnDataMissing` variant.

Rendered output is intentionally unchanged — I audited how many pages
this affects before deciding on the approach.

## Audit

Cross-checked all CSS property pages in `mdn/content` against the
currently published `mdn-data`: **562 property folders, 17 currently hit
this exact silent path** (`row-rule`, `row-rule-color`, `user-modify`,
`color-interpolation`, etc. — not just the `link-parameters` case from
the original report, which has since self-healed once mdn-data caught up).

Given that, returning `Err` instead of `Ok` would have changed the
*rendered page content* on all 17 (raw error text replacing the whole
formal-definition table) — a bigger and more visible regression than
fixing #821 alone. So this stays a soft flaw: same fallback UI, but now
recorded, instead of a hard build failure.

## Verified locally

Before (on `row-rule`, currently missing from mdn-data):


crates/rari-doc/src/issues.rs
```diff
@@ -337,6 +339,11 @@ pub enum DIssue {
         macro_name: Option<String>,
         href: Option<String>,
     },
+    MdnDataMissing {
```
LeoMcA: Let's not add a new variant, this is an error in a macro so we should use the `Macros` variant

crates/rari-doc/src/issues.rs
```diff
@@ -317,6 +318,7 @@ impl FromStr for IssueType {
             "redirected-link" => Self::RedirectedLink,
             "broken-link" => Self::BrokenLink,
             "ill-cased-link" => Self::IllCasedLink,
+            "mdn-data-missing" => Self::MdnDataMissing,
```
LeoMcA: Let's make this templ- to be consistent, perhaps move up the `match` too:

```suggestion
            "templ-mdn-data-missing" => Self::TemplMdnDataMissing,
```

LeoMcA (changes requested):

BHUVANSH855:
Done, moved it under the `Templ*` variants with the `templ-mdn-data-missing` source string, and switched it to reuse `DIssue::Macros` via `issue_source()` (same pattern as `TemplInvalidArg`/`TemplIllCasedArg`). Verified locally: shows up in `flaws.macros` as `TemplMdnDataMissing` with `macroName: "cssinfo"`, no other pages affected.

LeoMcA (approved):
Nice, clean - thank you!

PR #865, opened 2026-09-03
feat(templ): report which macro argument is at fault

caugner:
### Description

Name the macro and the offending argument in templ argument errors, and split the catch-all `must be provided` into three distinct messages:

- `must be provided`: nothing in that slot (`{{cssxref()}}`)
- `must not be empty`: a blank slot (`{{cssxref("")}}`, `{{cssxref(,"CSS")}}`)
- `could not be parsed`: an unparseable token (`{{htmlelement(「label」)}}`)

Reported under a new `templ-arg-error` source, so they land in the `macros` flaw bucket with `macroName` set instead of in `unknown`.

### Motivation

Make a templ argument flaw fixable from the report alone. `must be provided` said neither which argument was at fault nor which of three unrelated content bugs it was, each needing a different fix.

### Additional details

```
must be provided  ->  Macro cssxref argument 1 (name) must be provided
must be provided  ->  Macro domxref argument 1 (api_name) must not be empty
must be provided  ->  Macro htmlelement argument 1 (element_name) could not be parsed
must be a string  ->  Macro jsxref argument 3 (anchor) must be a string
```

A full build reports the same 17469 issues at identical positions as `main`; only these 13 messages change.

### DIFF
--- a/crates/rari-types/src/lib.rs
+++ b/crates/rari-types/src/lib.rs
@@ -25,6 +25,31 @@ pub enum ArgError {
     MustBeBool,
     #[error("must be provided")]
     MustBeProvided,
+    #[error("must not be empty")]
+    MustNotBeEmpty,
+    #[error("could not be parsed")]
+    MustBeParsable,
+    /// An [`ArgError`] annotated with the templ and the 1-based position and
+    /// name of the offending parameter.
+    #[error("{templ} argument {pos} ({name}) {source}")]
+    At {
+        templ: &'static str,
+        pos: usize,
+        name: &'static str,
+        #[source]
+        source: Box<ArgError>,
+    },
+}
+
+impl ArgError {
+    pub fn at(templ: &'static str, pos: usize, name: &'static str, source: ArgError) -> Self {
+        ArgError::At {
+            templ,
+            pos,
+            name,
+            source: Box::new(source),
+        }
+    }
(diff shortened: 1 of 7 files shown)


crates/rari-types/src/lib.rs
```diff
@@ -25,6 +25,31 @@ pub enum ArgError {
     MustBeBool,
     #[error("must be provided")]
     MustBeProvided,
+    #[error("must not be empty")]
+    MustNotBeEmpty,
+    #[error("could not be parsed")]
+    MustBeParsable,
+    /// An [`ArgError`] annotated with the templ and the 1-based position and
+    /// name of the offending parameter.
```
LeoMcA:
```suggestion
```

crates/rari-types/src/lib.rs
```diff
@@ -42,6 +67,14 @@ pub enum Arg {
     Bool(bool),
 }
 
+impl Arg {
+    /// Whether this is an empty string literal (`""`) or an omitted argument
+    /// (`{{foo(,"bar")}}`).
```
LeoMcA: Potentially misleading: reads like is_blank distinguishes between the two, regardless I think the method name alone is enough:

```suggestion
```

LeoMcA (approved):

PR #866, opened 2026-09-03
chore(templs): remove legacy URL repair from `QuickLinksWithSubpages`

caugner:
### Description

Remove `fix_broken_legacy_url` and its only call site in the `QuickLinksWithSubpages` macro.

- Delete `crates/rari-doc/src/templ/legacy.rs` and its `pub mod legacy;` declaration.
- `QuickLinksWithSubpages` now forwards its `url` argument to `listsubpages` unchanged.

### Motivation

The helper rewrote macro arguments that were missing a locale prefix or the `/docs/` segment into a canonical URL, logging a `fixed legacy url` warning. Once the remaining non-canonical arguments are fixed in translated-content, the repair is a no-op on every input, and removing it makes a future non-canonical argument fail page resolution and surface as a macro error rather than being silently repaired.

### Additional details

Across content and translated-content, `QuickLinksWithSubpages` is invoked with either a canonical `/<locale>/docs/...` URL or no argument at all, except for 21 occurrences of `{{QuickLinksWithSubPages("Web/Performance")}}` (es: 2, ko: 6, ru: 13). Those are exactly the 21 pages reported as `fixed legacy url` in a full build, and they are addressed by the linked translated-content PRs.

Arguments with a trailing slash (e.g. `/es/docs/Web/Progressive_web_apps/`) already took the identity branch of `fix_broken_legacy_url`, so nothing downstream relied on normalization happening here.

Depends on:

- mdn/translated-content#38218
- mdn/translated-content#38219
- mdn/translated-content#38220


LeoMcA (approved):
Merge after dependant translated-content PRs are merged

PR #868, opened 2026-09-03
feat(templ): report malformed macro syntax as its own flaw

caugner:
### Description

Report a macro with a syntax error as its own `templ-syntax-error` flaw instead of a misleading complaint about its arguments.

- Look the `args` node up by kind, so argument recovery is not thrown off by a preceding tree-sitter `ERROR` node.
- Record `MacroToken::malformed` from `Node::has_error`, and suppress the argument error a malformed macro would otherwise produce, since its recovered arguments are not trustworthy.

### Motivation

Point the author at the typo. The stray `(` in `{{cssxref(("color")}}` made rari drop every argument and render `must be provided` into the page, which says nothing about the actual mistake.

### Additional details

`{{cssxref(("color")}}` now renders the `color` link it was meant to produce and reports `Macro cssxref has invalid syntax`.

Against #865, a full build surfaces 26 malformed macros (23 previously silent) and replaces 3 misleading argument errors. Spot checks found no false positives: all are missing `)`, doubled `(`, or unescaped quotes. Rendered output is unchanged on every affected page except the one above.

**Depends on:** #865

### DIFF
--- a/crates/rari-doc/src/templ/parser.rs
+++ b/crates/rari-doc/src/templ/parser.rs
@@ -16,6 +16,10 @@ pub struct MacroToken {
     pub ident: String,
     pub pos: (usize, usize),
     pub args: Vec<Option<Arg>>,
+    /// Whether the macro contains a syntax error, e.g. a stray parenthesis in
+    /// `{{cssxref(("color")}}`. Its `args` are then whatever tree-sitter managed
+    /// to recover and should not be trusted.
+    pub malformed: bool,
 }
 
 fn from_node<'a>(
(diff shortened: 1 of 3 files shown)


crates/rari-doc/src/templ/parser.rs
```diff
@@ -16,6 +16,10 @@ pub struct MacroToken {
     pub ident: String,
     pub pos: (usize, usize),
     pub args: Vec<Option<Arg>>,
+    /// Whether the macro contains a syntax error, e.g. a stray parenthesis in
+    /// `{{cssxref(("color")}}`. Its `args` are then whatever tree-sitter managed
+    /// to recover and should not be trusted.
```
LeoMcA: "should not be trusted" implies we throw them out entirely to me, but we don't - I don't think we need the comment at all here, we document what we do with it inline, where it matters:

```suggestion
```

LeoMcA (approved):

crates/rari-doc/src/templ/render.rs
```diff
@@ -92,6 +92,12 @@ pub(crate) fn render(env: &RariEnv, input: &str, offset: usize) -> Result<Render
                     end_col = end_col
                 );
                 let _enter = span.enter();
+                if mac.malformed {
+                    warn!(
+                        source = "templ-syntax-error",
+                        "Macro {name} has invalid syntax"
```
caugner:
```suggestion
                        "Macro {name} has syntax error"
```

PR #870, opened 2026-09-03
fix(sidebars): skip sidebar macros on unrooted pages

caugner:
### Description

Skip sidebar macros on `conflicting/` and `orphaned/` pages.

- `invoke` returns an empty result for `TemplType::Sidebar` when the page slug is unrooted, before the templ function runs.
- `build_sidebars` bails out too, since a frontmatter entry naming a sidebar that is not a registered templ reaches `render_sidebar` without passing through `invoke`.
- Empty rendered sidebars are dropped in `build_content`, so `sidebarHTML` stays absent instead of becoming the empty document that `Html::parse_fragment("")` yields.
- New `is_unrooted(slug)` helper in `rari-doc/src/utils.rs`.

### Motivation

Unrooted slugs are not part of any document tree, so sidebar macros cannot produce anything meaningful there:

- `apiref` and `defaultapisidebar` fail with `InvalidSlugForX`.
- `jsref` derives an empty object name (`slug_to_object_name` returns `""` via `unwrap_or_default()`), then links to `/Web/JavaScript/Reference/Global_Objects/` with a trailing slash, which resolves through a redirect. It also renders an empty `<code></code>` label and a spurious `%Base%` inheritance section.
- Static, slug-independent sidebars render a tree that highlights nothing.

None of this is fixable by authors: the affected pages just call `{{JSRef}}` or `{{APIRef}}`.

This matches how `build_doc` already degrades these pages, with `no_indexing` and without parents or other translations.

### Additional details

Rebuilding the 335 `conflicting/` and `orphaned/` pages in `translated-content` goes from 463 to 407 issues. The 56 removed are exactly the sidebar ones, with none added:

| macro | issues removed | category |
| --- | --- | --- |
| `apiref` | 39 | `Invalid slug for templ/sidebar: …` |
| `jsref` | 16 | `templ-redirected-link` |
| `defaultapisidebar` | 1 | `Invalid slug for templ/sidebar: …` |

Rooted pages are unaffected. `Web/JavaScript/Reference/Global_Objects/Array/toString` (`jsref`), `Web/API/Window` (inline `apiref`) and `Web/HTML/Reference/Elements/div` (frontmatter YAML `htmlsidebar`) all still render their sidebars, and the remaining issues on those pages are identical to `main`.


crates/rari-doc/src/utils.rs
```diff
@@ -165,6 +165,11 @@ pub fn root_for_locale(locale: Locale) -> Result<&'static Path, EnvError> {
     }
 }
 
+/// Whether `slug` belongs to an unrooted (`conflicting/…` or `orphaned/…`) document.
```
LeoMcA:
```suggestion
```

LeoMcA (approved):

PR #875, opened 2026-09-04
feat(css-syntax): accept an optional CSS reference resolver

caugner:
### Description

Add an optional `CssRefResolver` callback to `render_formal_syntax`, letting the caller decide where a `<type>` or `<'property'>` node links, or that it shouldn't link at all.

- `CssRefKind` tells the resolver whether the node is a type or a property.
- A new `SyntaxRenderer::render_reference` helper replaces the two hand-rolled `format!` link builders in `render_node`.
- Passing `resolver: None` keeps the previous behavior, so this is a no-op on its own.

### Motivation

`render_node` hardcoded `Reference/Values/{slug}` and `Reference/Properties/{slug}` for every node. `css-syntax` has no access to the MDN page tree, so it cannot tell whether those pages exist, and the webref grammars reference many spec-internal productions that MDN does not document. Injecting a resolver lets `rari-doc` supply that knowledge without inverting the crate dependency.

Passing the node kind rather than just the slug means the resolver never has to re-derive the category from a URL or a naming convention.

### Additional details

Split out of #874 to keep the plumbing separate from the behavior change. All 20 existing `css-syntax` tests pass with unchanged expected HTML, which is what verifies the `None` path is a no-op; two new tests cover the `Some(resolver)` path for both node kinds.

Relates to #874, which uses this to fix 1857 broken formal-syntax links.

### DIFF
--- a/crates/css-syntax/src/syntax.rs
+++ b/crates/css-syntax/src/syntax.rs
@@ -258,12 +258,8 @@ impl CssRefKind {
     }
 }
 
-/// Maps a reference to the `Web/CSS/Reference/` sub-path documenting it (e.g.
-/// `url` -> `Values/url_value`), or `None` to render it unlinked. The slug is
-/// bare, but keeps a trailing `()` for function types.
-///
-/// Injected because `css-syntax` has no access to the MDN page tree; `None`
-/// links everything under the default category.
+/// Callback resolving a CSS type or property to an MDN reference path.
+/// If it returns `None` the term should be rendered without a link.
 pub type CssRefResolver<'a> = &'a dyn Fn(CssRefKind, &str) -> Option<String>;
 
 pub struct SyntaxRenderer<'a> {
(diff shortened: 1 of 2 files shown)


crates/css-syntax/src/syntax.rs
```diff
@@ -240,11 +240,38 @@ struct Term {
...
+        }
+    }
+}
+
+/// Maps a reference to the `Web/CSS/Reference/` sub-path documenting it (e.g.
+/// `url` -> `Values/url_value`), or `None` to render it unlinked. The slug is
+/// bare, but keeps a trailing `()` for function types.
+///
+/// Injected because `css-syntax` has no access to the MDN page tree; `None`
+/// links everything under the default category.
```
LeoMcA: Can we condense this, and avoid explaining past bugs:


```suggestion
/// A callback for resolving a CSS type or property to an MDN reference path.
/// The callback returns `None` when the term should be rendered without a link.
```

crates/css-syntax/src/syntax.rs
```diff
@@ -328,6 +355,25 @@ impl SyntaxRenderer<'_> {
         Ok(out)
     }
 
+    /// Render a `<type>` / `<'property'>` reference, linked only if the
+    /// resolver reports a page for it. Many webref productions are
+    /// spec-internal (`<number-token>`, `<any-value>`) and have none.
```
LeoMcA:
```suggestion
```

LeoMcA (approved):

crates/css-syntax/src/syntax.rs
caugner:
```suggestion
/// Callback resolving a CSS type or property to an MDN reference path.
/// If it returns `None` the term should be rendered without a link.
```

LeoMcA (approved):

PR #903, opened 2026-09-07
refactor(md): drop unused `tagfilter` handling

caugner:
### Description

Remove the unused comrak `tagfilter` handling from `rari-md`:

- drop the explicit `options.extension.tagfilter = false` assignment (already the default),
- drop the unreachable `tagfilter_block` branch in the HTML formatter,
- drop the now unused `tagfilter` and `tagfilter_block` helpers.

### Motivation

Ensure the build stays warning-free with comrak 0.55.0, which deprecates `Extension::tagfilter` (to be removed in 0.56.0). rari never enabled the extension, so this is a pure dead-code removal without behavior change.

Unblocks #901.


caugner:
See: https://github.com/mdn/rari/actions/runs/34128802983/job/101763694839?pr=903#step:10:20

> Took: 11.084189398s - 0 changes in 0 of 14633 files (0.0%)

🟢

LeoMcA (approved):

PR #904, opened 2026-09-07
refactor(issues): simplify href lookup in fix-flaws

caugner:
### Description

Simplify how `fix-flaws` locates an href in the raw markdown.

- Search forward from the start of the reported line with `find_non_prefix_match()`, and remove `actual_offset()` and the backward re-search anchored to its result.
- Extract `issue_to_suggestion()` and `build_candidates()` from `collect_suggestions()`, and replace `search_with_slug_fallback()` with an ordered list of `(search, replace)` candidates.
- Report the line number and search start offset in the "Could not locate" warning instead of an always-empty "searched region".
- Add a table-driven test for `build_candidates()`.

### Motivation

Make it easier to add further lookup candidates, and remove a misleading two-step search that did the same forward search twice.

### Additional details

`actual_offset()` never used the HTML column. It ran `find_non_prefix_match()` from the line start for the unescaped href or slug and returned the match end, which `collect_suggestions()` then re-found by searching backward to that offset. The `DisplayIssue` column refers to the rendered HTML rather than the markdown source, so the line is the only position information the search can rely on.

No behaviour change is intended. The existing `collect_suggestions()` tests pass unchanged.

Prepares #643.

### DIFF
--- a/crates/rari-tools/src/fix/issues.rs
+++ b/crates/rari-tools/src/fix/issues.rs
@@ -1044,9 +1044,9 @@ Some content here.
 
     #[test]
     fn test_repeated_href_on_same_line() {
-        // Regression test: the same href appearing multiple times on one line must each be fixed.
-        // Previously, actual_offset always found the *first* occurrence, producing identical
-        // SearchReplaceWithOffset entries that dedup collapsed to one.
+        // The same href appearing multiple times on one line must each be fixed. Without
+        // `next_search_from`, every issue would match the *first* occurrence, producing identical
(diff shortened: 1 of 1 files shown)


crates/rari-tools/src/fix/issues.rs
LeoMcA: Comment still refers to the removed `actual_offset`

LeoMcA (approved):

crates/rari-tools/src/fix/issues.rs
caugner: Good catch, thanks! Rewrote the comment to describe the behavior without naming the removed function in ea7b81ba.

PR #907, opened 2026-09-07
ci(pr-diff-companion): label PRs with build-diff page count

caugner:
### Description

Add a `pr-diff-companion.yml` workflow that labels PRs with the page count of their build diff:

- Triggered by `workflow_run` of the "Build Content and Diff" workflow.
- Downloads the `build-diff-stats` artifact of the triggering run and applies exactly one `diff: *` label (`diff: none`, `diff: 1-10 pages`, `diff: 11-100 pages`, `diff: 101-1000 pages`, `diff: 1000+ pages`).
- When pages changed, upserts a sticky comment with the numbers and a link to the `diff.html` report; otherwise updates an existing one to say no pages changed.
- Excludes that comment from the PR Review Companion's deploy comment selector so the two sticky comments do not overwrite each other.

### Motivation

Make "does this PR change any built output at all?" visible from the PR list. For dependency bumps, `diff: none` is effectively a green light.

### Additional details

The diff workflow runs on `pull_request` with a read-only token for Dependabot and fork PRs, so it cannot label itself, and a reusable workflow called from it cannot widen the token either. `workflow_run` runs in the base repository off the default branch with a full token.

Because the triggering run may come from a fork, the artifact is treated as untrusted:

- The PR number is resolved via the API from the trusted `workflow_run.head_sha`, filtering on `head.sha` (the commits-to-pulls endpoint also returns PRs that merely contain the commit, e.g. the upper PR of a stack). If no open PR has that head, a newer push superseded the run and the job exits without doing anything.
- The report artifact id comes from the runs API, not the artifact.
- Every number read from the JSON is type-checked with `jq` before it reaches a `gh` argument or the comment body.

`workflow_run` only fires from the default branch, so the job cannot be exercised on this PR. The five `diff: *` labels have to exist in the repository before this is merged; `gh pr edit --add-label` fails otherwise.

**Depends on:** #906


LeoMcA (approved):
Nice! I'm looking forward to this 🚀

PR #915, opened 2026-09-08
fix(serve): log asset requests at debug level

caugner:
### Description

Change `get_file_handler` in the dev server to log static asset requests via `tracing::debug!` instead of `tracing::info!`.

### Motivation

Prevent every CSS/JS/image fetch under `rari serve` from printing an info line and drowning out useful output.

Extracted from #390.

### DIFF
--- a/crates/rari-cli/serve.rs
+++ b/crates/rari-cli/serve.rs
@@ -94,7 +94,7 @@ async fn wrapped_fix_issues(
 
 async fn get_file_handler(req: Request) -> Result<Response, AppError> {
     let url = req.uri().path();
-    tracing::info!("(asset) {}", url);
+    tracing::debug!("(asset) {}", url);
     let UrlMeta {
         page_category,
         slug,


crates/rari-cli/serve.rs
```diff
@@ -94,7 +94,7 @@ async fn wrapped_fix_issues(
 
 async fn get_file_handler(req: Request) -> Result<Response, AppError> {
     let url = req.uri().path();
-    tracing::info!("(asset) {}", url);
+    tracing::debug!("(asset) {}", url);
```
LeoMcA, line 97: I don't think we ever emit `debug!` logs, we either have the level set to warning or off:

https://github.com/mdn/rari/blob/b3b94092d86c985321f88d66d36e83e06356b7fc/crates/rari-cli/main.rs#L441-L445

LeoMcA (changes requested):

crates/rari-cli/serve.rs
caugner: Oh, good point, I thought I had seen it being configurable, but maybe not.

Let's close for now, and if it comes up again, we can revisit.

PR #922, opened 2026-09-09
fix(npm): replace `extract-zip` with `yauzl`

caugner:
### Description

Replace `extract-zip` with `yauzl` in `rari-npm`. On Windows, `unpack` now extracts only the `rari.exe` entry from the release zip, skips any other entry, and rejects a `rari.exe` entry that is not a regular file (e.g. a symlink).

### Motivation

Resolve the two open Dependabot alerts against `extract-zip` (GHSA-jmr9-qjv8-65gv and GHSA-7pqw-9j4j-h8q3, both high severity, symlink path traversal / arbitrary file write). `extract-zip` has no patched version and no commits since 2022, so there is no upgrade path. `yauzl` (which `extract-zip` wraps) is actively maintained, only parses the archive, and leaves all filesystem writes to our code.

### Additional details

Verified on macOS with a copy of the new `extractZipEntry` helper:

| Case | Result |
|---|---|
| `rari-x86_64-pc-windows-msvc.zip` from the latest release | extracted `rari.exe` (17.2 MB) |
| Zip with a symlink entry named `rari.exe` pointing at `/etc/passwd` | rejected: not a regular file |
| Zip with an extra entry `other.txt` | skipped `other.txt`, extracted `rari.exe` |

Dropping `extract-zip` also removes its transitive dependencies (`get-stream`, `debug`, `yauzl@2`) from the lock file; `npm audit` reports no vulnerabilities.

### DIFF
--- a/rari-npm/lib/download.js
+++ b/rari-npm/lib/download.js
@@ -4,9 +4,10 @@ import * as https from "https";
 import { createWriteStream } from "fs";
 import { unlink, access, constants, mkdir, chmod } from "fs/promises";
 import path from "node:path";
+import { pipeline } from "node:stream/promises";
 
 import { x } from "tar";
-import extract from "extract-zip";
+import yauzl from "yauzl";
 
 import packageJson from "../package.json" with { type: "json" };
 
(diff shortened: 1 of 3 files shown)


caugner:
Closing in favor of https://github.com/mdn/rari/pull/925.

Issue #924, opened 2026-09-10
npm: publish per-platform packages


caugner:
### Problem

`@mdn/rari` downloads the binary from the GitHub release in a `postinstall` script. This needs network access to `api.github.com` at install time, is affected by rate limits and proxies, and pulls in `tar`, `extract-zip`, `https-proxy-agent` and `proxy-from-env` as runtime dependencies.

### Proposal

Follow the approach used by lefthook, esbuild and Biome:

- Publish one package per target (e.g. `@mdn/rari-darwin-arm64`, `@mdn/rari-linux-x64`) containing only the binary, restricted via `os` and `cpu` in its `package.json`.
- List them as `optionalDependencies` of `@mdn/rari`, pinned to the same version, so npm installs only the matching one.
- Resolve the binary at runtime with `require.resolve()` instead of `postinstall`, keeping the `rariBin` export unchanged for consumers like fred.
- Build the platform packages from the existing GitHub release assets in `publish-npm.yml`, so the release assets stay as they are.

caugner:
Resolved in @mdn/rari 1.0.0!

PR #925, opened 2026-09-10
feat(npm): load binary from platform package

caugner:
### Description

Update `@mdn/rari` to load its binary from a platform package pinned in `optionalDependencies`, replacing the `postinstall` download and its JavaScript dependencies.

Move type generation into the private `@mdn/rari-npm-tooling` package with its own lockfile. Add `RARI_BINARY_PATH` support and use it in rari's content-build workflow.

### Motivation

Let package managers install the binary without install scripts or GitHub downloads, and keep release tooling independent of unpublished platform versions.

### Additional details

Release-please updates the platform dependency pins. The exported `rariBin` API remains unchanged.

Rari's `_build.yml` keeps copying the test binary to the legacy path until content upgrades its npm wrapper.

Validated tarball installation and CLI execution, plus locked tooling installation, type generation, package preparation, and packing after a simulated version bump. Workflow lint passed.

Merge only after a release containing #928 has published the platform packages through the workflow.

Part of #924.

**Depends on:** #928.


.github/workflows/_build.yml
```diff
@@ -146,6 +146,9 @@ jobs:
         working-directory: mdn/rari
         run: |
           chmod +x target/release/rari
+          cp target/release/rari "$RUNNER_TEMP/rari"
```
LeoMcA: nit: could we set `RARI_BINARY_PATH` above and use it as the cp target here, for consistency?


```suggestion
          cp target/release/rari "$RARI_BINARY_PATH"
```

rari-npm/lib/index.js
```diff
@@ -1,6 +1,27 @@
...
+  const binary = process.platform === "win32" ? "rari.exe" : "rari";
+  try {
+    return require.resolve(`${PLATFORM_PACKAGE}/bin/${binary}`);
+  } catch (error) {
+    throw new Error(
+      `Could not find the rari binary for ${process.platform}-${process.arch}. ` +
+        `Make sure ${PLATFORM_PACKAGE} is installed: it is an optional dependency ` +
+        `of @mdn/rari, so it is missing when installing with --omit=optional, ` +
+        `when the platform is unsupported, or when the lockfile was generated on ` +
+        `another platform. Alternatively, set RARI_BINARY_PATH to a rari binary.`,
```
LeoMcA: I don't think "or when the lockfile was generated on another platform" is true, is it? Otherwise we wouldn't be able to use a checked-in lockfile across different platforms.

rari-npm/tooling/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@mdn/rari-npm-tooling",
+  "private": true,
+  "type": "module",
+  "devDependencies": {
+    "json-schema-to-typescript": "^16.0.0"
```
LeoMcA, line 6: We need to update the dependabot config to ensure this receives bumps

LeoMcA (approved):
Overall looks good, a couple of issues:

.github/workflows/_build.yml
```diff
@@ -168,6 +171,7 @@ jobs:
 
           # Increase GitHub API limit.
           GITHUB_TOKEN: ${{ github.token }}
+          RARI_BINARY_PATH: ${{ runner.temp }}/rari
```
caugner: ~~I think we shouldn't do this, because that environment variable isn't set in content, so the test would not reflect actual behavior.~~

rari-npm/lib/index.js
caugner: True, this should no longer be an issue with npm 11.8.0.

rari-npm/tooling/package.json
caugner: Yes, good catch!

.github/workflows/_build.yml
caugner: Thought about it again, and you're right.

.github/workflows/_build.yml
caugner: Addressed in 3f0d9168b876bda22d0a50a4bb346826dff0e4cf by defining `RARI_BINARY_PATH` at job scope and using it as the copy target.

rari-npm/lib/index.js
caugner: Addressed in 89d653242052bfd3fbe351aa487bbc1ba102e3fd by removing the obsolete lockfile guidance.

rari-npm/tooling/package.json
caugner: Addressed in 9d841ca7733f865509ae7db37b44ac967c3cb2d9 by adding weekly Dependabot updates for the tooling package.

.github/workflows/_build.yml
caugner: Updated in 3f0d9168b876bda22d0a50a4bb346826dff0e4cf. `RARI_BINARY_PATH` remains enabled and is now defined once at job scope for both the copy and content build.

PR #937, opened 2026-09-14
ci(publish-npm): support staged publishing

caugner:
### Description

Update `publish-npm.yml` to stage all seven npm packages with `npm stage publish`, reusing the binaries of an existing release.

- Add a `stage_tag` input selecting the release whose binaries to reuse.
- Assign `<version>-stage.<run-id>.<attempt>` to the wrapper, its platform pins, and the platform packages, and stage them under the `staging` dist-tag.
- Require staging to run on a branch and reject combining `stage_tag` with `recovery_tag`.
- Log a notice when the `package.json` version differs from `stage_tag`.
- Document staging, trusted publisher requirements, tarball testing, and cleanup in `rari-npm/README.md`.

### Motivation

Verify packaging and Trusted Publishing before cutting a release or switching consumers to platform packages.

### Additional details

Staging defaults to a dry run. Staged uploads use provenance and remain unavailable to normal installs until approved. Each attempt gets a unique version, so retries never collide with releases or pending stages, and the published-version skip only applies to releases. Release and recovery behavior is unchanged apart from switching the publish flags to bash arrays.

Each package's trusted publisher configuration on npmjs.com must allow `npm stage publish`; configurations created before September 2026 only allow `npm publish` by default.

Workflow inputs:

```diff
+stage_tag: Release tag whose binaries to reuse for staging (optional)
-publish: Actually publish (otherwise dry run)
+publish: Actually publish or stage (otherwise dry run)
```

Validated with actionlint, a local run of the version step against `rari-npm/package.json`, and `npm stage publish --access public --tag staging --dry-run` with the npm bundled in Node 24. OIDC authorization for staging requires an actual staging run.

Part of #924.

**Depends on:** #925.


.github/workflows/publish-npm.yml
```diff
@@ -83,9 +98,28 @@ jobs:
...
         run: |
-          TAG="v$(node -p 'require("./package.json").version')"
+          TAG="${STAGE_TAG:-v$(node -p 'require("./package.json").version')}"
           gh release download "$TAG" --repo "$GITHUB_REPOSITORY" --pattern 'rari-*' --dir "$RUNNER_TEMP/assets"
 
+      - name: Set staging version
+        if: ${{ inputs.stage_tag != '' }}
+        working-directory: ./rari-npm
+        run: |
+          node --input-type=module <<'JS'
```
LeoMcA, line 108: Can we use `actions/github-script` here instead?

rari-npm/README.md
```diff
@@ -87,3 +87,52 @@ Recovery verifies that the release tag matches the package version, that the
...
+gh workflow run publish-npm.yml --repo mdn/rari --ref my-branch \
+  -f stage_tag=v0.2.34 -f publish=false
+```
+
+After inspecting the dry run, repeat with `-f publish=true` to upload all seven
+packages using `npm stage publish` with provenance. They remain unavailable to
+normal installs until approved. This exercises npm authentication, which a dry
+run does not test. Each package's trusted publisher configuration on npmjs.com
+must allow `npm stage publish` (the "Allowed actions" setting); configurations
+created before September 2026 only allow `npm publish` by default.
```
LeoMcA, line 108: the before-september note feels like unnecessary detail, which will age quickly.

LeoMcA (changes requested):

caugner:
@LeoMcA I'm wondering if it's actually worth to add staged publishing given how it increases complexity of the publish workflow. What do you think: Drop or keep?

PR #938, opened 2026-09-14
fix(compat): report unknown BCD keys

### LINKED ISSUE #692: Broken browser-compat entries should cause flaw
Example: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color-interpolation-method#browser_compatibility

Image

The page lists two browser-compat entries:

```
browser-compat:
  - css.types.color.color-mix
  - css.types.color.color.display-p3-linear
```

However, `css.types.color.color.display-p3-linear` does not exist in BCD, and this does not show up as a flaw: https://caugner.github.io/mdn-flawless/#q=display-p3-linear

Noticed while writing https://github.com/mdn/rari/issues/691.

caugner:
### Description

Update compatibility macros to report missing BCD keys as flaws.

### Motivation

Make broken `browser-compat` entries visible in build issue reports.

### Additional details

Accept grouping keys containing compatibility data and preserve rendered sections.
Reuse cached BCD data only when both derived indexes are valid. Regenerate both if either is missing or malformed.

New flaw message:

```diff
+Unknown browser-compat entry: css.types.color.color.display-p3-linear
```

Validated with compatibility tests, Clippy, formatting checks, and a build of the reported example page.

Fixes #692.

Depends on:

- https://github.com/mdn/rari/pull/949


crates/rari-doc/src/templ/templs/compat.rs
```diff
@@ -41,11 +47,68 @@ pub fn webextallcompattables() -> Result<String, DocError> {
     Ok(out)
 }
 
+static BCD_KEYS: LazyLock<Option<HashSet<String>>> = LazyLock::new(|| {
```
LeoMcA, line 49: We already parse the full BCD data here:

https://github.com/mdn/rari/blob/75c7c4a5184bee2daefbaf05da9c25ed872fd49a/crates/rari-deps/src/bcd.rs#L12-L54

Could we move this, there, so we don't load and parse the full JSON again during the build?

LeoMcA (changes requested):

crates/rari-doc/src/templ/templs/compat.rs
caugner: Good catch, fixed via d6e3aebca3cee029a712279b37211fed490d5526.

crates/rari-deps/src/bcd.rs
```diff
@@ -11,7 +11,7 @@ use crate::npm::get_package;
 
 pub fn update_bcd(base_path: &Path) -> Result<(), DepsError> {
     if let Some(path) = get_package("@mdn/browser-compat-data", &deps().bcd, base_path)? {
-        extract_spec_urls(&path)?;
+        extract_data(&path)?;
```
LeoMcA: This won't run until the package updates or is deemed stale - perhaps that's something we're ok with? It'll self resolve with time, or a forced download.

LeoMcA (approved):
While there's still bcd data I believe this'll show a warning and just disable checking for browser-compat keys - I think this is ok, it'll self resolve.

If we did want to fix it we'd have to wire up some sort of migration generation, even when the package data isn't downloaded fresh, when the file is missing.

crates/rari-deps/src/bcd.rs
caugner: I think this is worth avoiding, so I opened https://github.com/mdn/rari/pull/949, and stacked this one on top of that.

caugner:
> **Build diff:** 64 changes in 64 of 14683 pages (0.4%)

These are the pages with unknown BCD keys! 🎉

PR #947, opened 2026-09-17
ci(dependabot): align Dependabot configuration

caugner:
### Description

Aligns the Dependabot configuration with the MDN defaults:

- Sets `open-pull-requests-limit: 10` on the `github-actions` entry and the `npm` entries for `/deps`, `/rari-npm` and `/rari-npm/tooling`
- Sets a 3-day cooldown on the `github-actions` entry, the `npm` entry for `/deps` and the `cargo` entry
- Groups minor and patch updates in the `npm` entry for `/deps` and the `cargo` entry
- Excludes `@mdn/*` from the groups in the `npm` entry for `/deps`

### Motivation

Ensures consistent dependency update configuration across all MDN repositories.

### Additional details

See also: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

Part of https://github.com/mdn/fred/issues/1445.


.github/dependabot.yml
```diff
@@ -6,8 +6,20 @@ updates:
...
+      cargo-prod:
+        dependency-type: production
+        update-types:
+          - minor
+          - patch
+      cargo-dev:
+        dependency-type: development
+        update-types:
+          - minor
+          - patch
```
caugner, line 22: IIUC Dependabot now correctly classifies 0.x bumps as major releases. If this is not the case, we can revert these groups, and I can raise it with the Dependabot PM.

PR #949, opened 2026-09-22
fix(deps): repair incomplete BCD cache

caugner:
### Description

Update BCD dependency handling to regenerate `spec_urls.json` when cached derived data is missing or invalid.

### Motivation

Ensure an interrupted dependency update self-heals on the next run.

### Additional details

`get_package` writes `last_check.json` before `spec_urls.json` is extracted. If rari is interrupted in between, the fresh check stamp skips the download on the next run, and the derived file stays missing until the next BCD version bump. Builds then log "Failed to load BCD spec URLs" and render empty specification tables.

`spec_urls.json` is still extracted after a fresh download. Otherwise it is now regenerated if it is missing or does not parse as a map of BCD keys to spec URL lists, which also covers files truncated by an interrupted write.

`get_package` now removes `last_check.json` before wiping a package directory, so an interrupted removal forces a re-download instead of leaving a fresh check stamp without `package/data.json`.

Related to #938.


crates/rari-deps/src/bcd.rs
```diff
@@ -10,9 +10,9 @@ use crate::error::DepsError;
 use crate::npm::get_package;
 
 pub fn update_bcd(base_path: &Path) -> Result<(), DepsError> {
-    if let Some(path) = get_package("@mdn/browser-compat-data", &deps().bcd, base_path)? {
-        extract_spec_urls(&path)?;
-    }
+    let package_path = base_path.join("@mdn/browser-compat-data");
+    get_package("@mdn/browser-compat-data", &deps().bcd, base_path)?;
+    ensure_spec_urls(&package_path)?;
```
LeoMcA: We're relying on the behaviour of `get_package` removing the old processed jsons for an updated package, rather than explicitly running the re-processing for an updated package - this makes me a little bit uncomfortable.

I think the intent is more readable here if we stick with the old `if let Some(path) ... { extract_spec_urls }` and add an `else { ensure_spec_urls }` branch.

It would also be nice to have a regression test which would catch this, testing: "does extract_spec_urls run when the package is stale"?

LeoMcA (changes requested):

crates/rari-deps/src/bcd.rs
caugner: Switched back to the explicit `if let Some(path) ... { extract_spec_urls } else { ensure_spec_urls }` shape, and added `test_update_bcd_data`, which covers both a downloaded package replacing valid spec URLs and a cached package regenerating missing ones.

LeoMcA (approved):

